PR #5, opened 2017-09-09
There doesn't seem to be a round_start for round_announce_match_start

zoidbergwill:


markus-wa:
`MatchStartedEvent` is currently sent on `begin_new_match`

I feel like it shouldn't be sent on two different GameEvents but maybe `round_announce_match_start` would make more sense as I've seen some sketchy behaviour with match starts.

Do you know which events occur when and how often exactly?

zoidbergwill:
Woah, on some of the test demos, both are raised multiple times, sometimes only one:

[match-start-events.log](https://gist.github.com/zoidbergwill/591b2b06aa11c53c19754d2ab6d05083)

I wonder if it changed over the years, since some of those demos are three or four years old.

I'm gonna try get one from each of the tournaments this weekend to test later.

markus-wa:
The test demos are professional matches. Each organizer has different ways of setting up their servers.

Back in the day many organizers just had admins manually running commands or scripts so that might be the reason some demos have these events multiple times (not sure how it is today).

The most reliable thing would be MM demos but I don't have any older ones so to test compatibility pro ones must do.

markus-wa:
I also think there might be demos where the score randomly resets to 0-0, the game restarts and then it goes back to where it was as well as other strange behaviour due to things like match medic.

So i think it's fine to send out MatchStartedEvent multiple times if it technically happened multiple times, just has to be consistent with the game events.

To make the tool easier to use in these scenariod I'd be open to having a feature to 'guess' the actual match start & side switches etc., suppressing additional events,
 (something like what I've put in the example in the README) - but that's a topic for another day.

zoidbergwill:
Oh, this is on the wrong branch of mine, will tidy it up.

PR #12, opened 2017-10-24
Add item events

zoidbergwill:
This is probably ready

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -263,3 +263,21 @@ type RankUpdateEvent struct {
 	WinCount   int
 	RankChange float32
 }
+
+// ItemEquipEvent signals an item was equipped
+type ItemEquipEvent struct {
+	Weapon common.Equipment
+	Player *common.Player
+}
+
+// ItemPickupEvent signals an item was bought or picked up.
+type ItemPickupEvent struct {
+	Weapon common.Equipment
+	Player *common.Player
+}
+
+// ItemDropEvent signals an item was dropped
+type ItemDropEvent struct {
+	Weapon common.Equipment
+	Player *common.Player
+}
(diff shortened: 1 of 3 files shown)


events/events.go
```diff
@@ -263,3 +263,21 @@ type RankUpdateEvent struct {
 	WinCount   int
 	RankChange float32
 }
+
+// ItemEquipEvent signals an item was equipped
+type ItemEquipEvent struct {
+	Weapon common.Equipment
```
markus-wa: Might wanna do this like the Nade events with nested structs.

packet_handlers.go
```diff
@@ -386,6 +386,33 @@ func (p *Parser) handleGameEvent(ge *msg.CSVCMsg_GameEvent) {
...
 
+	case "item_equip": // Equipped, I think
+		fallthrough
+	case "item_pickup": // Picked up or bought?
+		fallthrough
+	case "item_remove": // Dropped?
+		data = mapGameEventData(d, ge)
+		player := p.connectedPlayers[int(data["userid"].GetValShort())]
+		weapon := common.NewSkinEquipment(data["item"].GetValString(), "")
+
```
markus-wa: Nested structs as well, see buildNadeEvent() - but don't do be like me, that function really belongs into the events package. (I'll move the buildNadeEvent one as well now that I saw that).

markus-wa:
Thanks, looks good overall. Comments are really nitpicky, I know. If you want to I can merge it and fix it afterwards.

markus-wa:
After thinking about it my proposal is kinda shit, would be useless to add another struct just for all this. . .

Merging as is and sorry for me being me 🤐

PR #15, opened 2018-04-14
Expose copy of header

micvbang:
Hey,

I needed to use header information in my consumption of the API.

I suggest to return only a copy of the header since information in it is used during demo parsing, i.e. `Progress`.

Best,
Michael


micvbang:
Ah. I had actually completely missed the HeaderParsedEvent. Maybe this addition does not make sense anyway.

markus-wa:
No I think it's quite weird to use the event for header information tbh, not sure why I put it there.

I suggest removing the `HeaderParsedEvent` completely and returning a copy of the header in `ParseHeader()` (for ease of use). I would also like to keep your `Header()` function so the header doesnt need to be saved preemptively.

But in turn I'd suggest removi g the helper functions like `Parser.Map()`.

What do you think?

markus-wa:
I'm also wondering why `Parser.header` is a pointer. Will check later

micvbang:
Yes, I agree.

Since this is a library used by more people, I would suggest not removing the `HeaderParsedEvent`. It's likely that someone depends on it, and that removing it would break their code :)
As far as I understand, the way to change the API in Go libraries is to release them under a new URI. Please correct me if you think otherwise!

If you haven't done this before I get to work on it during next week, feel free do do the work on my PR. Also, if you have any code that you'd like reviewed, feel free to make a PR and assign it to me.

markus-wa:
Let me know if you think this looks good and I'll merge it.
I already took the liberty of rebasing it onto master.

micvbang:
Looks good!

PR #22, opened 2018-06-20
Add entity to NadeEvent

zyphie:


markus-wa:
Sorry, I misunderstood your request. I thought this was about adding the thrower (`common.Player`) onto the event, which is already there.

I think either exposing `Parser.entities` via a new function `Parser.Entities()` would make more sense; otherwise this new `Entity` field would have to be added to all player related events to be consistent and it would be somewhat redundant.

Then you could use `Parser.Entities()[thrower.EntityID]`.

If that makes sense to you could you change the PR to that?

markus-wa:
With `Parser.entities` exposed you could access the entities from anywhere else whenever you need them.

markus-wa:
In fact, I just noticed `Player.Entity` already exists :sweat_smile:

Does this work for you?

zyphie:
The entity I need is the entity of the nade, not the player. But you're probably right that it's better to have a generalized function to fetch entities, and then we can just add a field for entityID on the nade events. If you agree I can change the field etc to be for entityID instead.

markus-wa:
Actually, in that case it makes sense, sorry.

Could you rename it to `NadeEntity` though?

Exposing `Parser.entities` is probably still a good idea though, feel free to do that as well.

parser.go
```diff
@@ -79,6 +79,11 @@ func (p *Parser) GameState() *GameState {
 	return &p.gameState
 }
 
+// Entities returns the available entities
+func (p *Parser) Entities() *map[int]*st.Entity {
```
markus-wa: This should return a `map[int]*st.Entity`
Go maps are already references by default

zyphie: Ooops, my bad. Changed it.

markus-wa:
Thanks for the contribution 🙌

PR #23, opened 2018-06-21
Revert to use entity id instead of entity

zyphie:
The solution with using the entity doesn't really work. For some nades (e.g. molotov), the entity is deleted. If we have the ID, while we cannot access the entity, we can at least relate the different events (throw, explode, expire).

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -112,10 +112,10 @@ type NadeEventIf interface {
 // NadeEvent contains the common attributes of nade events. Dont register
 // handlers on this tho, you want NadeEventIf for that
 type NadeEvent struct {
-	NadeType common.EquipmentElement
-	Position r3.Vector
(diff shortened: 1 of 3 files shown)


markus-wa:
I wonder why we're so quick to destroy unused entities then.
Maybe this could be solved by not destroying the entity instead?

I'll look into it a bit more later, if not, this looks like the next best thing.

zyphie:
Entity IDs are not unique throughout a demo, though. I.e. multiple entities (that does not exist at the same time) can have the same ID. Wouldn't this pose a problem if we persist the entities?

markus-wa:
Good point, guess this'll have to do

markus-wa:
Could you rename it to `NadeEntityID` again tho? :)

PR #26, opened 2018-06-22
Added prop value to PropertyEntry

zyphie:
Added a way to get the current prop value. The current implementation only stores prop value for values that we have created a RegisterPropertyUpdateHandler on.

### DIFF
--- a/sendtables/entity.go
+++ b/sendtables/entity.go
@@ -127,6 +127,7 @@ type PropertyEntry struct {
 	index          int
 	entry          *FlattenedPropEntry
 	updateHandlers []PropertyUpdateHandler
+	value          PropValue
 }
 
 // Entry returns the underlying FlattenedPropEntry.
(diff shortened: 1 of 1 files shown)


zyphie:
Yeah, I agree.

micvbang:
This feature is definitely needed, as it can greatly simplify a lot of code.

I'm currently using this branch in my work to compute grenade (projectile) positions. 

I'm not sure (as in: this should be tested) if it would have a large impact on performance to store all property values. If it is the case that it impacts performance, a simple `dummyBind` method would help us, e.g.:

```
// dummyBind is used to bind an empty event handler to a property update, triggering an internal
// caching mechanism of the given property.
func dummyBind(e *st.Entity, p string) {
	e.FindProperty(p).RegisterPropertyUpdateHandler(func(val st.PropValue) {})
}
```

markus-wa:
Looking at travis performance doesn't seem to be an issue at all.

Will probably merge this later today.

markus-wa:
Thanks @zyphie

Note that I changed the name of the field & method so they are less stuttery (i.e. to avoid Prop[...].prop[...]).

PR #27, opened 2018-06-27
Dev

micvbang:
I made an attempt to add grenade projectile data.  I haven't tested it with a lot of different demos yet, but for the few demos that I did test with, it seems to work well.

Since it's in a works-on-my-machine state right now, I thought I'd get it on the dev branch ASAP so that you can get to play with it as well.

Note: since entity ids are being reused throughout demos, I had the idea of adding a `randomID` (int64) to the `GrenadeProjectile` struct. I think this probably makes sense to do for `Equipment` as well. It's easy to track entity creation events in this library, but much harder on the consumer side :slightly_smiling_face:

### DIFF
--- a/examples/bouncy-nades/bouncynades.go
+++ b/examples/bouncy-nades/bouncynades.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"fmt"
+	"image"
+	"image/color"
+	"log"
+	"os"
+
+	"github.com/golang/geo/r3"
+
+	draw2dimg "github.com/llgcode/draw2d/draw2dimg"
+
+	dem "github.com/markus-wa/demoinfocs-golang"
+	events "github.com/markus-wa/demoinfocs-golang/events"
+	ex "github.com/markus-wa/demoinfocs-golang/examples"
+	st "github.com/markus-wa/demoinfocs-golang/sendtables"
+)
+
+// Run like this: go run bouncynades.go -demo /path/to/demo.dem
+func main() {
+	f, err := os.Open(ex.DemoPathFromArgs())
+	checkError(err)
+	defer f.Close()
+
+	p := dem.NewParser(f)
+
+	_, err = p.ParseHeader()
+	checkError(err)
+
+	stp := p.SendTableParser()
+
+	// Initialize the graphic context on an RGBA image
+	dest := image.NewRGBA(image.Rect(0, 0, 1000, 1000))
+	gc := draw2dimg.NewGraphicContext(dest)
+
+	// Set some properties
+	gc.SetFillColor(color.RGBA{0x00, 0x00, 0x00, 0x00})
+	gc.SetStrokeColor(color.RGBA{0xff, 0x00, 0x00, 0xff})
+	gc.SetLineWidth(2)
+
+	lastBouncePositionsID := make(map[int]r3.Vector)
+	var lastEntityID int
+
+	first := true
+	stopped := false
+
+	drawNadeLine := func(entityID int, pos r3.Vector) {
+		if stopped {
+			return
+		}
+
+		lastPos, ok := lastBouncePositionsID[entityID]
+
+		x := (2000 + pos.X) / 5
+		y := (2000 - pos.Y) / 5
+		fmt.Println(entityID, pos.X, pos.Y, x, y)
+
+		if !ok {
+			if first {
+				first = false
+			} else {
+				gc.FillStroke()
+			}
+
+			gc.BeginPath()  // Initialize a new path
+			gc.MoveTo(x, y) // Move to a position to start the new path
+		} else {
+			if lastEntityID != entityID {
+				gc.FillStroke()
+
+				gc.BeginPath()                  // Initialize a new path
+				gc.MoveTo(lastPos.X, lastPos.Y) // Move to a position to start the new path
+			}
+			gc.LineTo(x, y)
+		}
+
+		lastEntityID = entityID
+		lastBouncePositionsID[entityID] = r3.Vector{X: x, Y: y}
+	}
+
+	p.RegisterEventHandler(func(events.RoundEndedEvent) {
+		stopped = true
+		p.Cancel()
+	})
+
+	p.RegisterEventHandler(func(e events.NadeEventIf) {
+		ne := e.Base()
+		drawNadeLine(ne.NadeEntityID, ne.Position)
+		delete(lastBouncePositionsID, ne.NadeEntityID)
+		if lastEntityID == ne.NadeEntityID {
+			lastEntityID = -1
+		}
+	})
+
+	p.RegisterEventHandler(func(events.DataTablesParsedEvent) {
+		projectileTypes := []string{"CBaseCSGrenadeProjectile", "CDecoyProjectile", "CMolotovProjectile", "CSmokeGrenadeProjectile"}
+
+		for _, pt := range projectileTypes {
+			stp.FindServerClassByName(pt).RegisterEntityCreatedHandler(func(e st.EntityCreatedEvent) {
+				ent := e.Entity
+				prop := ent.FindProperty("m_nBounces")
+				drawNadeLine(ent.ID, ent.Position())
+				prop.RegisterPropertyUpdateHandler(func(st.PropValue) {
+					drawNadeLine(ent.ID, ent.Position())
+				})
+			})
+		}
+	})
+
+	err = p.ParseToEnd()
+
+	gc.FillStroke()
+
+	// Save to file
+	draw2dimg.SaveToPngFile("bouncynades.png", dest)
+}
+
+func checkError(err error) {
+	if err != nil {
+		log.Fatal(err)
+	}
+}
--- a/examples/bouncy-nades/bouncynades_test.go
+++ b/examples/bouncy-nades/bouncynades_test.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+// Just make sure the example runs
+func TestBouncyNades(t *testing.T) {
+	os.Args = []string{"cmd", "-demo", "../../cs-demos/default.dem"}
+
+	main()
+}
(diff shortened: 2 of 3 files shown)


markus-wa:
looks interesting.

On a sidenote, I just pushed some conflicting code on dev 😅 77f803f
For example there's now a `Entity.Position()` function. We'll need to sort this somehow.

common/structs.go
```diff
@@ -100,12 +102,39 @@ type Equipment struct {
...
+}
+
 // Class returns the class of the equipment.
 // E.g. pistol, smg, heavy etc.
 func (e Equipment) Class() EquipmentClass {
 	return e.Weapon.Class()
 }
 
+// NewGrenadeProjectile creates a grenade projectile and sets.
+func NewGrenadeProjectile() *GrenadeProjectile {
```
markus-wa: I think I'd prefer a sequence on the parser or something like that. No collision chances and faster.

It probably wouldn't ever be a problem either way but seems somewhat cleaner to me.

common/constants.go
```diff
@@ -167,3 +173,15 @@ const (
...
+
+const (
+	ServerClassBaseGrenadeProjectile ServerClassID = 9
+	ServerClassDecoyProjectile       ServerClassID = 41
+	ServerClassMolotovProjectile     ServerClassID = 98
+	ServerClassSmokeProjectile       ServerClassID = 134
+)
+
+const (
+	EqModelFlash EquipmentModelIndex = 131
```
zyphie: These values are not constant between demos. Maybe get filenames from model index to figure out which nade.

If we'll be doing this for BaseGrenadeProjectile, I think we might as well do the same for {Decoy, Molotov, Smoke}Projectile. This is for backwards compatibility as these classes can have different values in older demos. E.g. I know that MolotovProjectile had class ID 97 previously.

datatables.go
```diff
@@ -334,3 +415,53 @@ func (p *Parser) bindWeapon(event st.EntityCreatedEvent) {
 		wepFix("_pist_p250", "_pist_cz_75", func() { eq.Weapon = common.EqCZ })
 	}
 }
+
+// dummyBind is used to bind an empty event handler to a property update, triggering an internal
+// caching mechanism of the given property.
+func dummyBind(e *st.Entity, p string) bool {
```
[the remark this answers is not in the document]
micvbang: Cool, will remove

datatables.go
```diff
@@ -334,3 +415,53 @@ func (p *Parser) bindWeapon(event st.EntityCreatedEvent) {
...
+	prop := e.FindProperty(p)
+	if prop == nil {
+		return false
+	}
+	prop.RegisterPropertyUpdateHandler(func(val st.PropValue) {})
+	return true
+}
+
+func getIntPropVal(e *st.Entity, propName string) int {
```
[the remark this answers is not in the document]
micvbang: Yeah, I think it should. I used it during debugging when I was trying to determine which values on the CBaseGrenade could be used to differentiate flashes and HEs from each other. Will probably remove this as well

datatables.go
```diff
@@ -334,3 +415,53 @@ func (p *Parser) bindWeapon(event st.EntityCreatedEvent) {
...
+	return prop.Value().IntVal
+}
+
+func getIntPropValAsFloat64(e *st.Entity, propName string) float64 {
+	return float64(getIntPropVal(e, propName))
+}
+
+// bindComputePosition binds sets up handlers that compute the world coordinate-position of objects
+// that are compressed during networking, using cell ids.
+func bindComputePosition(p *Parser, e *st.Entity, c *common.CellOrigin) bool {
```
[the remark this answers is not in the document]
micvbang: I'll rework to use this instead!

parser.go
```diff
@@ -102,6 +103,10 @@ func (p *Parser) Progress() float32 {
 	return float32(p.currentFrame) / float32(p.header.PlaybackFrames)
 }
 
+func (p *Parser) GrenadeProjectiles() map[int]*common.GrenadeProjectile {
```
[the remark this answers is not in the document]
micvbang: :+1:

common/constants.go
micvbang: Yeah, I thought that it might be the case that they change between demos. Do you happen to know where we can read this information?

zyphie: It is in the modelprecache. The model index can be used on the modelPreCache (on Parser) to get the relevant model path. I am guessing doing this once and persisting the mapping would be best?

So e.g. something like this pseudo code:

- Expose modelPreCache

```
func (p *Parser) ModelPreCache() []string {
	return p.modelPreCache
}
```

- Set up handlers to get the precache when it has been populated

```
p.RegisterEventHandler(func(e events.StringTableCreatedEvent) {
	if e.TableName == "modelprecache" {
		// Now we can find the model index for each nade based on name
		FindMdlIdx(p.ModelPreCache())
	}
})
```

- Find the index for the different nade projectiles, and then persist it somewhere?

```
func FindMdlIdx(m []string) {
	for i, n in := range m {
		if strings.Contains(m, "flashbang_dropped") {
			// i is the index of flashbang projectile
			// ... and so forth for every nade
		}
	}
}
```

common/structs.go
markus-wa: Yeah I think I commented on the wrong line, my bad.
And yeah just leave it for now, should be fine.

One other thing tho: I think the ID should be renamed to UniqueID or something, otherwise it's implying the library consumer isn't supposed to use it.

parser.go
```diff
@@ -102,6 +103,12 @@ func (p *Parser) Progress() float32 {
 	return float32(p.currentFrame) / float32(p.header.PlaybackFrames)
 }
 
+// GrenadeProjectiles returns a map with all grenade projectiles. The map contains only projectiles
+// that are currently in-flight, i.e. have been thrown but have yet to detonate.
+func (p *Parser) GrenadeProjectiles() map[int]*common.GrenadeProjectile {
```
markus-wa: I think it might be preferable to put this inside `Parser.gameState` if possible.

The parser is already fairly cluttered and it would fit pretty well there.

common/structs.go
```diff
@@ -100,12 +102,37 @@ type Equipment struct {
...
+	// @micvbang TODO: decide if we want to use random ints or a sequence from the parser here.
+	return &GrenadeProjectile{internalID: rand.Int63()}
+}
+
+// InternalID returns the internal id of the grenade, which is used to distinguish grenades that reuse another entity's entity id.
+func (g GrenadeProjectile) InternalID() int64 {
```
markus-wa: I think the ID should be renamed to UniqueID or something, otherwise it's implying the library consumer isn't supposed to use it.

common/structs.go
micvbang: Good point w.r.t. your edited comment. I'll try and see if I can find a better name for it. `UniqueID` is not bad, but I'll think about it some more :)

parser.go
micvbang: That's a good point!

common/structs.go
micvbang: Good point. I'll try and see if I can find a better name for it. `UniqueID` is not bad, but I'll think about it some more :)

common/structs.go
markus-wa: I moved that to a separate comment on the right line and yeah if you come up with something else that's good too.

common/constants.go
micvbang: That was an excellent description, thank you! I've given the implementation an attempt.

How did you know this? I'm pretty new to parsing these demo files, and I've spent a lot of time trying to find _any_ documentation on them. I've only had luck with stumbling over info on random forums, and various other implementations of this library.

common/structs.go
micvbang: I ended up with the suggested `UniqueID` name :slightly_smiling_face:

common/common.go
```diff
@@ -22,6 +22,17 @@ func init() {
 
 func initEqNameToWeapon() {
 	eqNameToWeapon = make(map[string]EquipmentElement)
+	// TODO: these seem to be unmapped
```
markus-wa: Yeah they are currently umapped since that should turn them into `EqUnknown` as it's the default value.

These weapons can't be obtained in-game with the exception of sensorgrenade, which could be obtained in some of the co-op missions, but we don't care about that.

It would be better to map them to it explicitly and then check if eqNameToWeapon contains a mapping and if not emit a warning. But we don't have easy access to dispatching events from this package right now.

See TODO in `MapEquipment`

common/common.go
```diff
@@ -60,7 +71,8 @@ func initEqNameToWeapon() {
 	eqNameToWeapon["scar20"] = EqScar20
 	eqNameToWeapon["sg556"] = EqSG556
 	eqNameToWeapon["smokegrenade"] = EqSmoke
-	eqNameToWeapon["ssg08"] = EqScout
+	eqNameToWeapon["smokegrenadeprojectile"] = EqSmoke
+	eqNameToWeapon["scout"] = EqScout
```
markus-wa: Is there a demo out there where this is actually necessary? This name is never used to my knowledge, maybe add it to the TODO above instead about unmapped weapons.

micvbang: It's not something that I really considered. I just happened to stumble upon and added it, I think. I'll remove it again.

common/common.go
micvbang: I'll remove these again and let this be handled in `MapEquipment` :slightly_smiling_face:

common/common.go
```diff
@@ -80,7 +80,7 @@ func initEqNameToWeapon() {
 	eqNameToWeapon["defuser"] = EqDefuseKit
 }
 
-func initEqEementToName() {
+func initEqElmentToName() {
```
markus-wa: Still a typo 😅 `Elment -> Element`

markus-wa:
Ehhhh, dammit. This wasn't supposed to happen . . .

I tried to fix the typo (which you already attempted to fix) and somehow i nuked the whole thing.

markus-wa:
Can you open a new one? The reopen button is locked here 😢

micvbang:
Sure thing :)

PR #30, opened 2018-07-05
Track player equipment

micvbang:
I've been playing with tracking players' current equipment.

In order to keep the `Player.CurrentEquipment` updated, I unfortunately had to use the `player_death` game event for resetting a player's equipment when he dies, `item_remove` for resetting items on forced round restarts, as well as using `bindGrenadeProjectiles` to track when grenades are removed from the player's inventory. I have been unable to find other ways to retrieve this information.

Once again, this is something that I haven't tested thoroughly on a large amount of demos, but on the demos that I tested it with, it seems to work reasonably well. I have noticed one bug, though, which is that resetting equipment on forced round restarts doesn't always work 100%. An example can be found in https://www.hltv.org/matches/2322780/liquid-vs-astralis-esl-pro-league-season-7-finals on round 16, where Magisk incorrectly, according to `Player.CurrentEquipment` holds both a USP and a cz.

I'm making it available here for you to play with, and hope that you guys can come with suggestions on how to improve it :slightly_smiling_face:

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -289,10 +289,6 @@ func (p *Parser) bindNewPlayer(playerEntity *st.Entity) {
 }
 
 func (p *Parser) bindWeapons() {
-	for i := 0; i < maxEntities; i++ {
-		p.weapons[i] = common.NewEquipment("")
-	}
-
 	for _, sc := range p.stParser.ServerClasses() {
 		for _, bc := range sc.BaseClasses {
 			switch bc.Name {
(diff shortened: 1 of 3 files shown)


micvbang:
I can see that there's a bug where the player's equipment doesn't get reset at half-time, as suggested by my comment here: https://github.com/markus-wa/demoinfocs-golang/pull/30/files#diff-4af3cd88b0d18d12411b7348b52d476cR409.

micvbang:
My idea w.r.t. resetting equipment on forced restarts doesn't seem to work; I can't find an event that only happens on forced restarts. Hmm.

markus-wa:
Good idea.

Apparently #31 is a thing - so we should probably replace `Player.Weapons` with what you are proposing.

datatables.go
```diff
@@ -328,34 +324,48 @@ func (p *Parser) bindGrenadeProjectiles(event st.EntityCreatedEvent) {
...
+		proj.Thrower = p.propValueToPlayer(val)
+
+		// The Owner of a grenade is not modified in CWeaponCSBase when the grenade is thrown. In order to keep
+		// Player.CurrentEquipment updated, we remove thrown grenades here.
+		// @micvbang FYI: once a grenade is thrown it becomes a grenade projectile. The entity ID of a grenade
+		// projectile is _not_ the same as the grenade that was thrown. I haven't been able to find a better
+		// way to identify the thrown grenade, other than simply looping over all CurrentEquipment.
+		for _, eq := range proj.Thrower.CurrentEquipment {
+			if eq.Weapon == proj.Weapon {
+				ammo := proj.Thrower.AmmoLeft[eq.AmmoType]
```
markus-wa: This is a problem for `cs-demos/valve_matchmaking.dem`

https://travis-ci.org/markus-wa/demoinfocs-golang/jobs/400343552#L612

micvbang: Ah. Without going deeper into it, I would bet that it has to do with the thrower not being set for fire grenades in some demos.

markus-wa:
I think I've actually fixed #31 (or attempted to anyway)

Could you check what's missing from the current implementation now that it works? Maybe we can just build on what's there.

The fix is on dev.

micvbang:
As far as I can tell, 1c41d30 fixes what I intended to fix with my PR while being much, much cleaner. I'll close this one!

PR #33, opened 2018-07-06
Add unique id to Equipment.

micvbang:
As I argued for GrenadeProjectiles, I find it very useful as a consumer of the library to have unique ids on grenade projectiles. I find the same to be the case for equipment :slightly_smiling_face:

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -291,10 +291,6 @@ func (p *Parser) bindNewPlayer(playerEntity *st.Entity) {
 }
 
 func (p *Parser) bindWeapons() {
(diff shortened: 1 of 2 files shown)


datatables.go
```diff
@@ -374,7 +370,8 @@ func (p *Parser) bindGrenadeProjectiles(event st.EntityCreatedEvent) {
 }
 
 func (p *Parser) bindWeapon(event st.EntityCreatedEvent) {
-	eq := &p.weapons[event.Entity.ID]
+	eq := common.NewEquipment("")
```
markus-wa: Good idea to do this here, actually

PR #34, opened 2018-07-06
 datatables: fix bug where entries in Parser.weapons are not updated correctly.

micvbang:
I didn't catch this bug that I introduced when I was testing bb14185.

The effect is that the data on each weapon is never updated. This PR should fix it.


micvbang:
Alright, I'll keep that in mind!

PR #37, opened 2018-07-26
metadata: add and use generic map translation and scaling.

micvbang:
Here is a suggestion to make the example a bit more generic.

As you can see, I added new maps (from https://github.com/zoidbergwill/csgo-overviews) and the translations and scaling for each one, as given in the repository.

As we talked about on gitter, it looks like convex hull is a neat way to give simple 2d information to users. 

I think we should try to add an interface to currently active infernos in the `GameState`. I hope to get time to look at this tomorrow. I'll probably write you on gitter! :slightly_smiling_face:


metadata/maps.go
```diff
@@ -0,0 +1,123 @@
...
+		Name:  "de_train",
+		PosX:  -2477,
+		PosY:  2392,
+		Scale: 4.7,
+	}
+)
+
+// MapNameToMap takes a map name and returns a Map instance.
+// If the map does not exist, an empty Map will be returned.
+func MapNameToMap(name string) Map {
```
markus-wa: And I think `MapNameToMap()` should just be a `map[string]Map`.

markus-wa:
Awesome stuff btw! Great idea 😄

markus-wa:
Oh and maybe add a 'courtesy of @zoidbergwill/csgo-overviews' thing into a README.

Gopkg.lock
```diff
@@ -3,80 +3,106 @@
...
+    "math/fixed",
   ]
+  pruneopts = "UT"
   revision = "c73c2afc3b812cdd6385de5a50616511c4a3d458"
 
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "cda22bd3cd5cfcaa6e9fab1e1f64ed6670859d06ea62922ac4cb42f7e4e701f7"
+  input-imports = [
```
markus-wa: Do you know what happened to the Gopkg.lock?

New dep version that I wasn't using yet maybe?

metadata/maps.go
micvbang: Yeah, that's a good point!

Gopkg.lock
micvbang: I honestly don't know - this is the first that I've ever used it :slightly_smiling_face:

micvbang:
Something like this? :slightly_smiling_face:

markus-wa:
Perfect

PR #41, opened 2018-08-27
Add GameState.Bomb()

micvbang:
I was looking into tracking the bomb's position, and it seems like it's something that could be useful to include in the library.

As you can see from the TODOs, I found a slight problem with the current interface of `Property.OnUpdate`. It seems like it could be useful to return a `HandlerIdentifier`, as is already being done in the event dispatcher.

I'm not sure where you would like this to go, but since I'm actively using branch v1.0.0 now, I branched out from here and pointed the PR here as well :slightly_smiling_face:

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -69,83 +69,40 @@ func (p *Parser) bindEntities() {
 }
 
 func (p *Parser) bindBomb() {
-	p.gameState.bomb = common.Bomb{}
 	bomb := &p.gameState.bomb
 
 	// Track bomb when it is not held by a player
-	p.stParser.ServerClasses().FindByName("CC4").OnEntityCreated(func(bomb *st.Entity) {
+	scDroppedC4 := p.stParser.ServerClasses().FindByName("CC4")
+	scDroppedC4.OnEntityCreated(func(bomb *st.Entity) {
 		bomb.OnPositionUpdate(func(pos r3.Vector) {
-			p.gameState.bomb.Player = nil // Bomb only has a position when not held by a player
-			p.gameState.bomb.Position = pos
+			// Bomb only has a position when not held by a player
+			p.gameState.bomb.Carrier = nil
+
+			p.gameState.bomb.LastOnGroundPosition = pos
 		})
 	})
 
 	// Track bomb when it has been planted
 	scPlantedC4 := p.stParser.ServerClasses().FindByName("CPlantedC4")
 	scPlantedC4.OnEntityCreated(func(bombEntity *st.Entity) {
-		p.gameState.bomb.Player = nil // Player can't hold the bomb when it has been planted
-		bomb.Position = bombEntity.Position()
+		// Player can't hold the bomb when it has been planted
+		p.gameState.bomb.Carrier = nil
+
+		bomb.LastOnGroundPosition = bombEntity.Position()
 	})
 
 	// Track bomb when it is being held by a player
 	scPlayerC4 := p.stParser.ServerClasses().FindByName("CC4")
 	scPlayerC4.OnEntityCreated(func(bombEntity *st.Entity) {
+		// TODO: @micvbang please document why this was introduced
 		wep := p.equipmentMapping[scPlayerC4]
 		if wep != common.EqBomb {
 			return
 		}
 
-		// TODO: fix this hack to "unregister" OnUpdate handlers
-		handlerRegistered := make(map[int]bool)
-
-		bombEntity.FindProperty("m_hPrevOwner").OnUpdate(func(val st.PropertyValue) {
-			ownerEntityID := val.IntVal & entityHandleIndexMask
-			handlerRegistered[ownerEntityID] = false
-		})
-
 		bombEntity.FindProperty("m_hOwner").OnUpdate(func(val st.PropertyValue) {
-			// TODO: for performance, it would be nice to be able to unregister the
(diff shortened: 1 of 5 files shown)


markus-wa:
Cool idea.

Regarding the OnUpdate issue: unfortunately i fear it's too big of a memory & performance overhead to track the update handlers in entities since there are so many of them constantly being created and then deleted compared to the value gained.

But I'm sure we'll find a way to get this in with or without the current hack.

datatables.go
```diff
@@ -65,6 +66,81 @@ func (p *Parser) bindEntities() {
...
+func (p *Parser) bindBomb() {
+	bombState := p.gameState.bombState
+
+	// Track bomb when it's not held by a player
+	p.stParser.ServerClasses().FindByName("CC4").OnEntityCreated(func(bomb *st.Entity) {
+		bombState.Position = bomb.Position()
+
+		bomb.OnPositionUpdate(func(pos r3.Vector) {
+			bombState.Player = nil // Bomb only has a position when not held by a player
+			bombState.Position = bomb.Position()
```
markus-wa: Why not `bombState.Position = pos`?

datatables.go
```diff
@@ -65,6 +66,81 @@ func (p *Parser) bindEntities() {
...
 	p.bindWeapons()
+	p.bindBomb()
+}
+
+func (p *Parser) bindBomb() {
+	bombState := p.gameState.bombState
+
+	// Track bomb when it's not held by a player
+	p.stParser.ServerClasses().FindByName("CC4").OnEntityCreated(func(bomb *st.Entity) {
+		bombState.Position = bomb.Position()
```
markus-wa: Shouldn't be neccessary, the `OnPositionUpdate` handler should be called with the initial value.

datatables.go
```diff
@@ -65,6 +66,81 @@ func (p *Parser) bindEntities() {
...
+
+		bomb.OnPositionUpdate(func(pos r3.Vector) {
+			bombState.Player = nil // Bomb only has a position when not held by a player
+			bombState.Position = bomb.Position()
+		})
+		bomb.OnDestroy(func() {
+			bombState.Position = bomb.Position()
```
markus-wa: What does this do?

datatables.go
```diff
@@ -65,6 +66,81 @@ func (p *Parser) bindEntities() {
...
+				return
+			}
+			handlerRegistered[ownerEntityID] = true
+
+			updateBombPosition := func() {
+				if !handlerRegistered[ownerEntityID] {
+					return
+				}
+
+				if _, hasBomb := owner.RawWeapons[bombEntity.ID()]; !hasBomb {
```
markus-wa: Maybe comment why this is needed (I guess it can't be assumed to always be true?)

game_state.go
```diff
@@ -15,6 +15,7 @@ type GameState struct {
 	grenadeProjectiles map[int]*common.GrenadeProjectile // Maps entity-IDs to active nade-projectiles. That's grenades that have been thrown, but have not yet detonated.
 	infernos           map[int]*common.Inferno           // Maps entity-IDs to active infernos.
 	entities           map[int]*st.Entity                // Maps entity IDs to entities
+	bombState          *common.BombState
```
markus-wa: I think it's nicer to take the reference when needed rather than storing it as pointer here.

micvbang: Excellent points and very nice catches w.r.t. code that was convoluted/not necessary!

micvbang:
Excellent points and very nice catches w.r.t. code that was convoluted/not necessary!

In writing a comment for the hasBomb check, I even learned something new about the behavior of the game.

I've updated accordingly :slightly_smiling_face:

micvbang:
Thank you!

Good comments. I've updated accordingly :slightly_smiling_face:

markus-wa:
I made some changes and added a function `Position()` that takes the position from the `Carrier` if set or otherwise from `LastOnGroundPosition`. That way we don't have register a ton of update handlers,

I also removed the check if the player has the bomb since I didn't have the entityID. What I did instead is check if the player held a `EqBomb` weapon, but that never got hit, so I removed it as well:
image

Please verify if that check is still necessary with the new code.

Let me know what your opinion is on this and any recommendations you have regarding the approach and code.

datatables.go
```diff
@@ -65,6 +65,46 @@ func (p *Parser) bindEntities() {
...
+		// Player can't hold the bomb when it has been planted
+		p.gameState.bomb.Carrier = nil
+
+		bomb.LastOnGroundPosition = bombEntity.Position()
+	})
+
+	// Track bomb when it is being held by a player
+	scPlayerC4 := p.stParser.ServerClasses().FindByName("CC4")
+	scPlayerC4.OnEntityCreated(func(bombEntity *st.Entity) {
+		// TODO: @micvbang please document why this was introduced
```
markus-wa: I saw this was getting hit by some demos in the test set but I'm not sure why it's needed.

Do you remember what it's for?

micvbang:
Excellent changes! This is much simpler than what we had before.
You're doing a great job of reviewing changes and keeping the library simple and performant. Good job!

With the demos that I previously tested, I see no differences in the output. I say we go with these improvements :slightly_smiling_face:

markus-wa:
Thank you :slightly_smiling_face:

Your contributions are also very much appreciated!

Issue #42, opened 2018-08-28
Smoke grenade OnDestroy event not triggered as expected


micvbang:
**Describe the bug**
In datatables.bindGrenadeProjectiles we assume that the projectile entity destroy-event is triggered _every time_ the entity no longer appears in the game. I still haven't found out _why_, but this does not seem to be the case. It looks to me like projectile destruction event is not triggered on smokes (more entities?) that do not expire normally, but are "removed" because the round has ended.

**To Reproduce**
I found the bug in round 4 of de_nuke in the following game: https://www.hltv.org/matches/2322780/liquid-vs-astralis-esl-pro-league-season-7-finals

```
package main

import (
	"log"
	"os"
	"sync"

	dem "github.com/markus-wa/demoinfocs-golang"
	"github.com/markus-wa/demoinfocs-golang/events"
)

func main() {
	f, err := os.Open("liquid-vs-astralis-m2-nuke.dem")
	if err != nil {
		log.Fatal(err)
	}

	p := dem.NewParser(f)
	_, err = p.ParseHeader()
	if err != nil {
		log.Fatal(err)
	}

	wg := &sync.WaitGroup{}

	roundNum := 0
	p.RegisterEventHandler(func(_ events.RoundOfficiallyEndedEvent) {
		roundNum++
	})

	p.RegisterEventHandler(func(_ events.RoundStartedEvent) {
		log.Printf("Round %+v just started. We expect there to be no projectiles. There are: %+v\n", roundNum, len(p.GameState().GrenadeProjectiles()))
	})

	err = p.ParseToEnd()
	if err != nil {
		log.Fatal(err)
	}

	wg.Wait()
}

```

**Expected behavior**
I expect GameState().GrenadeProjectiles() to only contain _active_ grenade projectiles, even those that did not expire normally

**Library version**
v1.0.0

A quick (and dirty) way to solve the problem is to simply reset the gameState.grenadeProjectiles map on RoundOfficiallyEnded event, like so (`game_events.go`):

```
...
case "round_officially_ended": // Round ended. . . probably the event where you get teleported to the spawn (=> You can still walk around between round_end and this?)
	for _, proj := range p.gameState.grenadeProjectiles {
		delete(p.gameState.grenadeProjectiles, proj.EntityID)
	}
	p.eventDispatcher.Dispatch(events.RoundOfficiallyEndedEvent{})
case ...
```

markus-wa:
[...]

PR #43, opened 2018-08-29
Issue 42 - Grenade entities not destoryed

markus-wa:
See #42 

@micvbang please check if this helps.

I had to do some more complicated code than you suggested but this should catch everything, hopefully.

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -285,11 +285,7 @@ func (p *Parser) bindGrenadeProjectiles(entity *st.Entity) {
 	})
 
 	entity.OnDestroy(func() {
-		p.eventDispatcher.Dispatch(events.NadeProjectileDestroyedEvent{
-			Projectile: proj,
-		})
-
-		delete(p.gameState.grenadeProjectiles, entityID)
+		p.nadeProjectileDestroyed(proj)
 	})
 
 	entity.FindProperty("m_nModelIndex").OnUpdate(func(val st.PropertyValue) {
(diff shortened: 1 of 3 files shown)


datatables.go
```diff
@@ -285,11 +285,14 @@ func (p *Parser) bindGrenadeProjectiles(entity *st.Entity) {
 	})
 
 	entity.OnDestroy(func() {
-		p.eventDispatcher.Dispatch(events.NadeProjectileDestroyedEvent{
-			Projectile: proj,
-		})
+		// Issue #42
+		// Sometimes entities are destoyed even after round_officially_ended, sometimes not
+		// Due to that we need to check if we already deleted it, otherwise we get a nil pointer dereference
```
micvbang: Consider re-reading this comment :slightly_smiling_face:

datatables.go
```diff
@@ -383,10 +395,14 @@ func (p *Parser) bindNewInferno(entity *st.Entity) {
 	})
 
 	entity.OnDestroy(func() {
-		p.eventDispatcher.Dispatch(events.InfernoExpiredEvent{
-			Inferno: inf,
-		})
-		delete(p.gameState.infernos, entityID)
+		// Issue #42
+		// Sometimes entities are destoyed even after round_officially_ended, sometimes not
+		// Due to that we need to check if we already deleted it, otherwise we get a nil pointer dereference
```
micvbang: Same as above comment

game_events.go
```diff
@@ -79,7 +79,22 @@ func (p *Parser) handleGameEvent(ge *msg.CSVCMsg_GameEvent) {
...
 
-	case "round_officially_ended": // Round ended. . . probably the event where you get teleported to the spawn (=> You can still walk around between round_end and this?)
+	case "round_officially_ended": // The event after which you get teleported to the spawn (=> You can still walk around between round_end and this event)
+		// Issue #42
+		// Sometimes grenades & infernos aren't deleted / destroyed via entity-updates at the end of the round,
+		// so we need to do it here for those that weren't.
+		//
+		// We're not deleting them from entitites though as that's supposed to be as close to the actual demo data as possible.
+		// We're also not using Entity.Destroy() because it would - in some cases - be called twice on the same entity
+		// and it's supposed to be called when the demo actually says so (same case as with GameState.entities).
```
micvbang: I quite agree with your reasoning here.

datatables.go
markus-wa: Sorry, I'm not quite following you here? Could you elaborate?

micvbang: Maybe it's just me, but I find that it's too general for what it actually covers.

Maybe something like the following?
> Sometimes grenade projectile entities are automatically destroyed after round_officially_ended, and sometimes they are not.
> The below check accommodates for the times they are not

markus-wa: > The below check accommodates for the times they are not.
Actually the opposite! :slightly_smiling_face: but I've changed it to hopefully be more clear.

Issue #45, opened 2018-09-04
v1.0.0 Event naming 


micvbang:
I was just going through some code and realized that the Event-suffix on all events in the `events` package is unnecessary and somewhat stuttery since it's already specified by the package name.

What do you think?

markus-wa:
I agree.

Feel free to submit a PR if you have time, otherwise I'll try to get to it over the weekend

PR #46, opened 2018-09-08
Clean up event naming

markus-wa:
See #45

Cc @micvbang - Any feedback appreciated.

I also decided to switch from past to present tense, removing many 'ed' suffixes, except where necessary so the events aren't confused with something else.

E.g. if it were `BombDefuse` instead of `BombDefused` it could be mistaken for the start of defusing.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -1,4 +1,8 @@
 // Package events contains all events that can be sent out from demoinfocs.Parser.
+//
+// Events are generally named in the tense that fits the best for each event.
+// E.g. SmokeExpired is in the past tense because it's sent out when the smoke has completely faded away
+// while SmokeStart is in the present tense because it's sent out when the smoke starts to bloom.
 package events
 
 import (
--- a/events/events.go
+++ b/events/events.go
@@ -179,9 +179,8 @@ type SmokeStart struct {
 	GrenadeEvent
 }
 
-// SmokeExpire signals the end of a smoke (fade). Not sure if this means
-// it started to fade, completely faded or something in between.
-type SmokeExpire struct {
+// SmokeExpired signals that a smoke as completely faded away.
+type SmokeExpired struct {
 	GrenadeEvent
 }
 
@@ -190,8 +189,8 @@ type FireGrenadeStart struct {
 	GrenadeEvent
 }
 
-// FireGrenadeExpire signals the end of a molly/incendiary.
-type FireGrenadeExpire struct {
+// FireGrenadeExpired signals that all fires of a molly/incendiary have extinguished.
+type FireGrenadeExpired struct {
 	GrenadeEvent
 }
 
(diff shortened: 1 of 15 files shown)


micvbang:
Looks good!

I'm not entirely sure about the change to present tense. Hmm. Maybe I've just gotten too used to the past tense, but I feel like it sounds better.

Is it on purpose that `GrenadeEvent` wasn't renamed to `Grenade`? I think that would make sense :slightly_smiling_face: Along with `GrenadeEventIf` -> `GrenadeIf`

markus-wa:
I had two main reasons for favoring present tense.
1. At the point where the event is fired, it's 'happening' in that instant, not in the past (possibly even just *starting* to happen).
    1. The event may arguably be in the future, if we're in the middle of the demo.
2. The demo's internal events are present-tense as well

Examples:

1.
`SmokeExpire`: "The smoke is expiring now"
I assume that this signals the game to start fading away the smoke

`WeaponFire`: "A weapon is BEING fired" (even tho it has 'ed' here it is still present tense.
This would probably instruct the game to *start* rendering the muzzle flash and playing weapon sounds.

2.
```
GameEvent: smokegrenade_expired Data: map[userid:248 entityid:347 x:948.243 y:266.19666 z:1616.0312]
GameEvent: weapon_fire Data: map[weapon:weapon_m4a1 silenced:false userid:249]
GameEvent: player_footstep Data: map[userid:254]
GameEvent: weapon_fire Data: map[silenced:false userid:233 weapon:weapon_ak47]
GameEvent: weapon_fire Data: map[silenced:false userid:249 weapon:weapon_m4a1]
GameEvent: weapon_fire Data: map[weapon:weapon_ak47 silenced:false userid:233]
GameEvent: player_footstep Data: map[userid:236]
GameEvent: player_footstep Data: map[userid:254]
```

On the `GrenadeEvent`, that's on purpose, so it wouldn't be confused with being an actual grenade for some reason. I'd prefer to keep it like this. Same with the `BombEvent`

micvbang:
Yeah, I appreciate those points.

Haha, that's funny. I always assumed the opposite of you, that grenades were gone by the time the expired event was thrown.

I just did a very scientific experiment :microscope:, staring at a replay of a smoke for a while. From what I can tell, the smoke actually begins to fade _before_ the SmokeExpire event is thrown, and is completely gone by the time it's thrown.

Maybe it makes sense to differentiate tense used for each event, based on what actually happens in-game?

Btw., the assumption about gunfire seems to be accurate!

markus-wa:
Very interesting, thanks for analyzing that!

In that case I agree, the tense should fit to what is correct for each case.

markus-wa:
I changed some more stuff, let me know if you see anything else
Otherwise I think I'll merge it this WE.

events/events.go
```diff
@@ -1,4 +1,8 @@
 // Package events contains all events that can be sent out from demoinfocs.Parser.
+//
+// Events are generally named in the tense that fits the best for each event.
+// E.g. SmokeExpired is in the past tense because it's sent out when the smoke has completely faded away
+// while SmokeStart is in the present tense because it's sent out when the smoke starts to bloom.
```
micvbang: This is excellent!

micvbang:
Sorry, I completely forgot about this PR. Once again great work, looking forward to using it!

markus-wa:
No worries 😄 and thanks!

PR #52, opened 2018-09-29
Fix players being skipped in processStringTable

markus-wa:
Due to an error when porting StatsHelix's demoinfocs, players were
skipped if there was a 'gap' between the string-table entries.

So if the string-table's userdata at index 3 was empty for example,
players at index 4 and above would be ignored.

### DIFF
--- a/stringtables.go
+++ b/stringtables.go
@@ -194,7 +194,7 @@ func (p *Parser) processStringTable(tab *msg.CSVCMsg_CreateStringTable) {
 		}
 
 		if len(userdat) == 0 {
-			break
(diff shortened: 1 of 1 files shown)


markus-wa:
@viSual42 - can you verify this fix works?

If you've been using the `v1.0.0` beta versions please check on the branch `fix-missing-players-v1` instead.

viSual42:
perfect ... yep works with my demos :D 
thanks!

markus-wa:
Cool, thanks for the confirmation 🙂

PR #55, opened 2018-10-16
Add ScoreUpdated event

### LINKED ISSUE #54: RoundEnd event not called every round
## Describe the bug
Hey Markus!

This is yet another I-found-a-bug-but-did-not-yet-investigate situation :slightly_smiling_face: 

I've found that in some demos the `RoundEnd` event is not triggered every round.

Since the internal score reported by the library is correct, at least `m_scoreTotal` is being called as we would expect.

## To Reproduce
Run the following code on this match: https://www.hltv.org/matches/2327343/mibr-vs-azio-ecs-season-6-north-america

```go
package main

import (
	"fmt"
	"log"
	"os"

	dem "github.com/markus-wa/demoinfocs-golang"
	"github.com/markus-wa/demoinfocs-golang/events"
	ex "github.com/markus-wa/demoinfocs-golang/examples"
)

func main() {
	f, err := os.Open(ex.DemoPathFromArgs())
	checkError(err)
	defer f.Close()

	p := dem.NewParser(f)

	_, err = p.ParseHeader()
	checkError(err)

	numRoundEnd := 1
	p.RegisterEventHandler(func(ev events.RoundEnd) {
		numRoundEnd++
	})

	numRoundEndOfficial := 1
	p.RegisterEventHandler(func(ev events.RoundEndOfficial) {
		numRoundEndOfficial++
	})

	err = p.ParseToEnd()
	checkError(err)

	fmt.Println("")
	fmt.Printf("Number of 'round end' events: %d\n", numRoundEnd)
	fmt.Printf("Number of 'round official end' events: %d\n", numRoundEndOfficial)
	fmt.Printf("Final score: %d:%d\n", p.GameState().TeamCounterTerrorists().Score(), p.GameState().TeamTerrorists().Score())
}

func checkError(err error) {
	if err != nil {
		log.Fatal(err)
	}
}
```

## Expected behavior
I would expect to see 26 `RoundEnd` events, as reported by HLTV

## Library version
I'm getting these results on `9916e2b4`

micvbang:
Hey!

Here's a draft of the implementation for the `ScoreUpdated` event.

Because of circular dependencies, I had to move `TeamState` to common.

What do you think? :slightly_smiling_face: 

Fixes #54

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -115,7 +115,7 @@ func (p *Parser) bindTeamScores() {
 		entity.FindProperty("m_szTeamname").OnUpdate(func(val st.PropertyValue) {
 			team := val.StringVal
 
-			var s *TeamState
+			var s *common.TeamState
 
 			switch team {
 			case "CT":
@@ -133,16 +133,23 @@ func (p *Parser) bindTeamScores() {
 
 			if s != nil {
 				// Set values that were already updated
-				s.id = teamID
-				s.clanName = clanName
-				s.flag = flagImage
-				s.score = score
+				s.ID = teamID
+				s.ClanName = clanName
+				s.Flag = flagImage
+				s.Score = score
 
 				// Register direct updates for the future
 				// Except for teamId, it doesn't change; players swap teams instead
-				entity.BindProperty("m_szClanTeamname", &s.clanName, st.ValTypeString)
-				entity.BindProperty("m_szTeamFlagImage", &s.flag, st.ValTypeString)
-				entity.BindProperty("m_scoreTotal", &s.score, st.ValTypeInt)
+				entity.BindProperty("m_szClanTeamname", &s.ClanName, st.ValTypeString)
+				entity.BindProperty("m_szTeamFlagImage", &s.Flag, st.ValTypeString)
+
+				entity.FindProperty("m_scoreTotal").OnUpdate(func(val st.PropertyValue) {
+					s.Score = val.IntVal
+
+					p.eventDispatcher.Dispatch(events.ScoreUpdated{
+						TeamState: s,
+					})
+				})
 			}
 		})
 	})
--- a/game_state.go
+++ b/game_state.go
@@ -8,8 +8,8 @@ import (
 // GameState contains all game-state relevant information.
 type GameState struct {
 	ingameTick         int
-	tState             TeamState
-	ctState            TeamState
+	tState             common.TeamState
+	ctState            common.TeamState
 	playersByUserID    map[int]*common.Player            // Maps user-IDs to players
 	playersByEntityID  map[int]*common.Player            // Maps entity-IDs to players
 	grenadeProjectiles map[int]*common.GrenadeProjectile // Maps entity-IDs to active nade-projectiles. That's grenades that have been thrown, but have not yet detonated.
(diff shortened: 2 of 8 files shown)


datatables.go
```diff
@@ -133,16 +133,23 @@ func (p *Parser) bindTeamScores() {
...
-				entity.BindProperty("m_scoreTotal", &s.score, st.ValTypeInt)
+				entity.BindProperty("m_szClanTeamname", &s.ClanName, st.ValTypeString)
+				entity.BindProperty("m_szTeamFlagImage", &s.Flag, st.ValTypeString)
+
+				entity.FindProperty("m_scoreTotal").OnUpdate(func(val st.PropertyValue) {
+					s.Score = val.IntVal
+
+					p.eventDispatcher.Dispatch(events.ScoreUpdated{
+						TeamState: s,
+					})
```
micvbang: Hmm. I wonder if we want to dispatch this event on 0-0 score as well?

markus-wa:
Cool! I like it overall. Unfortunate that it's api-breaking but i guess we'll have to live with that.

I should probably check if there's other structs etc that might need to be moved for future-proofness

But I can see travis isn't too pleased yet. Seems to be examples having bad imports now ;)

datatables.go
markus-wa: Good point. We should also consider the case of a forced score-reset though (so we don't go from 1-0 to 0-1 without an event for 0-0) e.g. if the match wasn't live yet etc. etc.

micvbang:
Yeah, that really is a shame. I guess it can go into the v1.0.0 branch, but obviously it means that it's not possible to back-port. At least as-is. 
If back-porting is important, I guess one way to do that would be to not move `TeamState`, but instead create a new type to be used for the new `ResultChanged` event.

markus-wa:
I don't think a backport is neccessary. I only plan to backport bug fixes to `0.x`.
It's just that I had never planned for this many `v1.0.0-beta.N` versions 😅.

Anyway, I don't think it's an issue, was just thinking out loud.
Overall it makes sense for `TeamState` to be located in `common` since it's similar to `Player`

common/teamstate.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+// TeamState contains a team's ID, score, clan name & country flag.
+type TeamState struct {
```
markus-wa: I think this can be moved to `common.go` being as small as it is.

game_state.go
```diff
@@ -89,39 +89,6 @@ func newGameState() GameState {
...
-	id       int
-	score    int
-	clanName string
-	flag     string
-}
-
-// ID returns the team-ID.
-//
-// This stays the same even after switching sides.
-func (ts TeamState) ID() int {
```
markus-wa: Please re-add this comment 🙂

game_state.go
```diff
@@ -89,39 +89,6 @@ func newGameState() GameState {
...
-// ClanName returns the team's clan name.
-func (ts TeamState) ClanName() string {
-	return ts.clanName
-}
-
-// Flag returns the team's country flag.
-// E.g. DE, FR, etc.
-//
-// Watch out, in some demos this is upper-case and in some lower-case.
-func (ts TeamState) Flag() string {
```
markus-wa: Also re-add this comment

datatables.go
markus-wa: I just checked and I think this can be ignored. `OnUpdate()` isn't triggered for the initial `0` value.

game_state.go
micvbang: Ah, very good catch!

game_state.go
micvbang: Likewise!

micvbang:
Haha, yeah, I can see that the number of releases is a bit higher than one would have guessed! :sweat_smile: 

All of your suggestions should have been fixed now!

markus-wa:
Awesome, thanks!

PR #56, opened 2018-10-17
game_events: begin dispatching events after any 'player_' event

micvbang:
Hey Markus!

I just found that in map 1 (dust2) of this demo https://www.hltv.org/matches/2322780/liquid-vs-astralis-esl-pro-league-season-7-finals, the demo recording is started _after_ players have connected. 

This means that no events are dispatched until `playersByUserID` has length > 0. `playersByUserID` is updated _later_ than `Parser.rawPlayers`, and does not necessarily contain any players before interesting events (e.g. `MatchStart`), although there are players on the server.

Here's a suggestion for a fix

### DIFF
--- a/game_events.go
+++ b/game_events.go
@@ -36,11 +36,6 @@ func (p *Parser) handleGameEvent(ge *msg.CSVCMsg_GameEvent) {
 
 	debugGameEvent(d, ge)
 
-	// Ignore events before players are connected to speed things up
-	if len(p.gameState.playersByUserID) == 0 && d.Name != "player_connect" {
-		return
-	}
-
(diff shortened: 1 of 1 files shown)


micvbang:
Other ways to fix this include 1) removing the optimization, 2) using `strings.HasPrefix(d.Name, "player_")` instead of `d.Name == "player_connected"`

markus-wa:
Hmm I wonder why I added this. Over all the demos in `cs-demos/set` it's only skipping 14 events.

I'd say we better remove it completely.

micvbang:
Thank you!

It has been removed now :slightly_smiling_face:

Issue #40, opened 2018-08-27
Exposing vendored types to user, e.g. in OnPositionUpdate(r3.Vector)


micvbang:
It may be that I'm just unaware of how this can be circumvented, but I'm having trouble using the callbacks that require the usage of vendored types, e.g. `Entity.OnPositionUpdate(r3.Vector)`. 

It makes sense that I can't just use the type as `r3.Vector` in my source, as it may be different from the vendored version. And the compiler won't let me import vendored packages, which I'm not absolutely certain that I understand why wouldn't work. Anyway, unless there's some way out of this that I can't see, I think we need to look at the types from vendored code used in callbacks to publicly exported code :slightly_smiling_face: 

The most obvious solution that I can come up with is to create the simplest possible wrappers for these types, e.g. where `r3.Vector` is used, one could create the type `Vector` and use it like so:

```
type Vector struct {
    r3.Vector
}

// example of internal use in demoinfocs-golang
func test(v r3.Vector) Vector {
    return Vector{v}
}
```

markus-wa:
Interesting. I'm wondering if this might also be resolvable by switching from dep to vgo modules.

I'll experiment a bit with that and see where it takes me.

micvbang:
Sounds good!

markus-wa:
I got it working by either using dep or vgo. I'd suggest using dep for now since vgo still isn't supported by default until go 1.12

Let me know if this helps.

### Code

```
package main

import (
	"fmt"
	"os"

	r3 "github.com/golang/geo/r3"

	dem "github.com/markus-wa/demoinfocs-golang"
)

func main() {
	f, _ := os.Open("asdf")
	p := dem.NewParser(f)

	p.GameState().Entities()[0].OnPositionUpdate(func(pos r3.Vector) {
		fmt.Println(pos)
	})
}
```

### dep

`Gopkg.toml`
```toml
[[constraint]]
  branch = "master"
  name = "github.com/golang/geo"

[[constraint]]
  name = "github.com/markus-wa/demoinfocs-golang"
  branch = "v1.0.0"

[prune]
  go-tests = true
  unused-packages = true
```

### vgo

`go.mod`
```
module demoinfocs-consumer

require (
	github.com/golang/geo v0.0.0-20180826223333-635502111454
	github.com/markus-wa/demoinfocs-golang v0.0.0-20180826223333-710bc4b6373601
)
```

markus-wa:
[...]

PR #57, opened 2018-10-18
Make Parser mockable

micvbang:
Hey Markus!

While implementing regression tests for handling different edge cases with regard to e.g. missing `RoundEnd` events, I found that it is incredibly useful to be able to mock `Parser`. 
Instead of having to have an actual demo for every regression test, mocking `Parser` allows us to recreate and look only at the essential events that cause the bug.

This PR is a draft of how this change could look.

Below is an example of a regression test I wrote using a mock implementation of `Parser`.
(I left out the mock itself since it's pretty long. It's incredibly simple, though. Tell me if you want a copy :slightly_smiling_face: )

```go
func TestRegressionMultipleFreezetimes(t *testing.T) {
	const expectedNumRounds = 2

	roundCh := parseRounds(common.DemoHeader{
		PlaybackTime:   150,
		PlaybackFrames: 1500,
	}, func(p *parserMock) {
		p.addEventDispatch(events.MatchStart{})
		p.addEventDispatch(events.RoundStart{})
		p.addEventDispatch(events.ScoreUpdated{})
		p.addEventDispatch(events.RoundEndOfficial{})

		p.addEventDispatch(events.RoundStart{})
		p.addEventDispatch(events.RoundEnd{})
		p.addEventDispatch(events.ScoreUpdated{})
		p.addEventDispatch(events.RoundEndOfficial{})
	})

	numRounds := 0
	for range roundCh {
		numRounds++
	}

	if expectedNumRounds != numRounds {
		log.Fatalf("Expected %d number of rounds, got %d", expectedNumRounds, numRounds)
	}
}
```

I understand that this is a pretty huge change, but I do believe that it is well worth the cost!

### DIFF
--- a/demoinfocs_test.go
+++ b/demoinfocs_test.go
@@ -14,8 +14,6 @@ import (
 	"time"
 
 	proto "github.com/gogo/protobuf/proto"
-	dispatch "github.com/markus-wa/godispatch"
-
 	dem "github.com/markus-wa/demoinfocs-golang"
 	common "github.com/markus-wa/demoinfocs-golang/common"
 	events "github.com/markus-wa/demoinfocs-golang/events"
@@ -134,7 +132,7 @@ func TestDemoInfoCs(t *testing.T) {
 	})
 
 	// Net-message stuff
-	var netTickHandlerID dispatch.HandlerIdentifier
+	var netTickHandlerID dem.HandlerIdentifier
 	netTickHandlerID = p.RegisterNetMessageHandler(func(tick *msg.CNETMsg_Tick) {
 		fmt.Println("Net-message tick handled, unregistering - tick:", tick.Tick)
 		p.UnregisterNetMessageHandler(netTickHandlerID)
@@ -246,7 +244,7 @@ func TestCancelParseToEnd(t *testing.T) {
 	maxTicks := 100
 	var tix int
 
-	var handlerID dispatch.HandlerIdentifier
+	var handlerID dem.HandlerIdentifier
 	handlerID = p.RegisterEventHandler(func(events.TickDone) {
 		tix++
 		if tix == maxTicks {
--- a/demoinfocs_test.go
+++ b/demoinfocs_test.go
@@ -14,6 +14,8 @@ import (
 	"time"
 
 	proto "github.com/gogo/protobuf/proto"
+	dispatch "github.com/markus-wa/godispatch"
+
 	dem "github.com/markus-wa/demoinfocs-golang"
 	common "github.com/markus-wa/demoinfocs-golang/common"
 	events "github.com/markus-wa/demoinfocs-golang/events"
@@ -132,7 +134,7 @@ func TestDemoInfoCs(t *testing.T) {
 	})
 
 	// Net-message stuff
-	var netTickHandlerID dem.HandlerIdentifier
+	var netTickHandlerID dispatch.HandlerIdentifier
 	netTickHandlerID = p.RegisterNetMessageHandler(func(tick *msg.CNETMsg_Tick) {
 		fmt.Println("Net-message tick handled, unregistering - tick:", tick.Tick)
 		p.UnregisterNetMessageHandler(netTickHandlerID)
@@ -244,7 +246,7 @@ func TestCancelParseToEnd(t *testing.T) {
 	maxTicks := 100
 	var tix int
 
-	var handlerID dem.HandlerIdentifier
+	var handlerID dispatch.HandlerIdentifier
 	handlerID = p.RegisterEventHandler(func(events.TickDone) {
 		tix++
 		if tix == maxTicks {
--- a/demopacket.go
+++ b/demopacket.go
@@ -12,7 +12,7 @@ import (
 
 // NetMessageCreator creates additional net-messages to be dispatched to net-message handlers.
 //
-// See also: ParserConfig.AdditionalNetMessageCreators & Parser.RegisterNetMessageHandler()
+// See also: ParserConfig.AdditionalNetMessageCreators & parser.RegisterNetMessageHandler()
 type NetMessageCreator func() proto.Message
 
 var byteSlicePool = sync.Pool{
@@ -22,7 +22,7 @@ var byteSlicePool = sync.Pool{
 	},
 }
 
-func (p *Parser) parsePacket() {
+func (p *parser) parsePacket() {
 	// Booooring
 	// 152 bytes CommandInfo, 4 bytes SeqNrIn, 4 bytes SeqNrOut
 	// See at the bottom what the CommandInfo would contain if you are interested.
--- a/demopacket.go
+++ b/demopacket.go
@@ -5,60 +5,15 @@ import (
 	"sync"
 
 	proto "github.com/gogo/protobuf/proto"
+
 	events "github.com/markus-wa/demoinfocs-golang/events"
 	msg "github.com/markus-wa/demoinfocs-golang/msg"
 )
 
 // NetMessageCreator creates additional net-messages to be dispatched to net-message handlers.
 //
-// See also: ParserConfig.AdditionalNetMessageCreators & parser.RegisterNetMessageHandler()
+// See also: ParserConfig.AdditionalNetMessageCreators & Parser.RegisterNetMessageHandler()
 type NetMessageCreator func() proto.Message
-
-func NewNetMessageFunc(t interface{}) (int, func() proto.Message) {
-	switch t.(type) {
-	case msg.CNETMsg_NOP:
-		return int(msg.NET_Messages_net_NOP), func() proto.Message {
-			return new(msg.CNETMsg_NOP)
-		}
-	case msg.CNETMsg_Disconnect:
-		return int(msg.NET_Messages_net_Disconnect), func() proto.Message {
-			return new(msg.CNETMsg_Disconnect)
-		}
-	case msg.CNETMsg_File:
-		return int(msg.NET_Messages_net_File), func() proto.Message {
-			return new(msg.CNETMsg_File)
-		}
-	case msg.CNETMsg_SplitScreenUser:
-		return int(msg.NET_Messages_net_SplitScreenUser), func() proto.Message {
-			return new(msg.CNETMsg_SplitScreenUser)
-		}
-	case msg.CNETMsg_Tick:
-		return int(msg.NET_Messages_net_Tick), func() proto.Message {
-			return new(msg.CNETMsg_Tick)
-		}
-	case msg.CNETMsg_StringCmd:
-		return int(msg.NET_Messages_net_StringCmd), func() proto.Message {
-			return new(msg.CNETMsg_StringCmd)
-		}
-	case msg.CNETMsg_SetConVar:
-		return int(msg.NET_Messages_net_SetConVar), func() proto.Message {
-			return new(msg.CNETMsg_SetConVar)
-		}
-	case msg.CNETMsg_SignonState:
-		return int(msg.NET_Messages_net_SignonState), func() proto.Message {
-			return new(msg.CNETMsg_SignonState)
-		}
-	case msg.CNETMsg_PlayerAvatarData:
-		return int(msg.NET_Messages_net_PlayerAvatarData), func() proto.Message {
-			return new(msg.CNETMsg_PlayerAvatarData)
-		}
-
-	default:
-		return -1, func() proto.Message {
-			return nil
-		}
-	}
-}
 
 var byteSlicePool = sync.Pool{
 	New: func() interface{} {
--- a/demopacket.go
+++ b/demopacket.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 
 	proto "github.com/gogo/protobuf/proto"
-
 	events "github.com/markus-wa/demoinfocs-golang/events"
 	msg "github.com/markus-wa/demoinfocs-golang/msg"
 )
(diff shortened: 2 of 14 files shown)


markus-wa:
Thanks Michael!

I'll look into it a bit more over the coming days but I think it sounds like a worthwhile change.

markus-wa:
Btw. Travis isn't happy for now, package `fuzzy` again.

https://travis-ci.org/markus-wa/demoinfocs-golang/builds/443220312

markus-wa:
Oh and yeah if you could put up a gist for the mock that'd be super neat!

Would be nice to have a page on the Wiki or an example project explaining how to do this.

parser.go
```diff
@@ -156,46 +173,49 @@ Returns a identifier with which the handler can be removed via UnregisterNetMess
...
 	}
 }
 
-// NewParser creates a new Parser with the default configuration.
+// NewParser creates a new parser with the default configuration.
 // The demostream io.Reader (e.g. os.File or bytes.Reader) must provide demo data in the '.DEM' format.
 //
 // See also: NewCustomParser() & DefaultParserConfig
-func NewParser(demostream io.Reader) *Parser {
+func NewParser(demostream io.Reader) *parser {
```
markus-wa: I think this should return the interface

```suggestion
func NewParser(demostream io.Reader) Parser {
```

parser.go
```diff
@@ -220,19 +240,19 @@ type ParserConfig struct {
...
 	MsgQueueBufferSize: -1,
 }
 
-// NewParserWithConfig returns a new Parser with a custom configuration.
+// NewParserWithConfig returns a new parser with a custom configuration.
 //
 // See also: NewParser() & ParserConfig
-func NewParserWithConfig(demostream io.Reader, config ParserConfig) *Parser {
-	var p Parser
+func NewParserWithConfig(demostream io.Reader, config ParserConfig) *parser {
```
markus-wa: As above

```suggestion
func NewParserWithConfig(demostream io.Reader, config ParserConfig) Parser {
```

parser.go
```diff
@@ -220,19 +240,19 @@ type ParserConfig struct {
 //
 // See also: package fuzzy for existing emitters with fuzzy-logic that depends on the demo-type.
 type EventEmitter interface {
-	Register(parser *Parser, eventDispatcher func(event interface{}))
+	Register(parser *parser, eventDispatcher func(event interface{}))
```
markus-wa: This should probably take the interface as arg

```suggestion
	Register(parser Parser, eventDispatcher func(event interface{}))
```

parser.go
```diff
@@ -6,17 +6,16 @@ import (
 	"time"
 
 	r3 "github.com/golang/geo/r3"
-	dp "github.com/markus-wa/godispatch"
-
 	bit "github.com/markus-wa/demoinfocs-golang/bitread"
 	common "github.com/markus-wa/demoinfocs-golang/common"
 	msg "github.com/markus-wa/demoinfocs-golang/msg"
 	st "github.com/markus-wa/demoinfocs-golang/sendtables"
+	dp "github.com/markus-wa/godispatch"
```
markus-wa: Move this up again. The idea is to have all packages of this repo in one group 🙂

examples/net-messages/README.md
```diff
@@ -49,9 +50,9 @@ cfg.AdditionalNetMessageCreators = map[int]dem.NetMessageCreator{
 
 ## Registering net-message handlers
 
-To register a handler for net-messages `Parser.RegisterNetMessageHandler()` can be used.
+To register a handler for net-messages `parser.RegisterNetMessageHandler()` can be used.
```
markus-wa: upper-case

```suggestion
To register a handler for net-messages `Parser.RegisterNetMessageHandler()` can be used.
```

examples/entities/README.md
```diff
@@ -798,12 +800,13 @@ ServerClass: id=202 name=CWeaponAWP
...
 		m_iBurstShotsRemaining
 ```
+
 </details>
 
 ## Registering entity-creation & property-update handlers
 
 Registering the entity-creation handlers needs to be done after the DataTablesParsedEvent has been dispatched.
-Before that the server-classes won't be available in `Parser.ServerClasses()`.
+Before that the server-classes won't be available in `parser.ServerClasses()`.
```
markus-wa: upper-case

```suggestion
Before that the server-classes won't be available in `Parser.ServerClasses()`.
```

doc.go
```diff
@@ -2,7 +2,7 @@
 Package demoinfocs provides a demo parser for the game Counter-Strike: Global Offensive.
 It is based on the official demoinfogo tool by Valve as well as Stats Helix's demoinfo.
 
-A good entry point to using the library is the Parser type.
+A good entry point to using the library is the parser type.
```
markus-wa: upper-case

```suggestion
A good entry point to using the library is the Parser interface.
```

demoinfocs_test.go
```diff
@@ -151,7 +151,7 @@ func TestDemoInfoCs(t *testing.T) {
 			p.Cancel()
 			timer.Reset(time.Second * 1)
 			<-timer.C
-			t.Fatal("Parser locked up for more than one second after cancellation")
+			t.Fatal("parser locked up for more than one second after cancellation")
```
markus-wa: Revert to upper-case

```suggestion
			t.Fatal("Parser locked up for more than one second after cancellation")
```

demoinfocs_test.go
```diff
@@ -163,7 +163,7 @@ func TestDemoInfoCs(t *testing.T) {
 			t.Fatal(err)
 		}
 		if !ok {
-			t.Fatalf("Parser reported end of demo after less than %d frames", frameByFrameCount)
+			t.Fatalf("parser reported end of demo after less than %d frames", frameByFrameCount)
```
markus-wa: Revert to upper-case

```suggestion
			t.Fatalf("Parser reported end of demo after less than %d frames", frameByFrameCount)
```

demopacket.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // NetMessageCreator creates additional net-messages to be dispatched to net-message handlers.
 //
-// See also: ParserConfig.AdditionalNetMessageCreators & Parser.RegisterNetMessageHandler()
+// See also: ParserConfig.AdditionalNetMessageCreators & parser.RegisterNetMessageHandler()
```
markus-wa: Should stay upper-case otherwise we reference the private type as only public types are in the godoc.

```suggestion
// See also: ParserConfig.AdditionalNetMessageCreators & Parser.RegisterNetMessageHandler()
```

events/events.go
```diff
@@ -1,4 +1,4 @@
-// Package events contains all events that can be sent out from demoinfocs.Parser.
+// Package events contains all events that can be sent out from demoinfocs.parser.
```
markus-wa: upper-case

```suggestion
// Package events contains all events that can be sent out from demoinfocs.Parser.
```

events/events.go
```diff
@@ -392,11 +392,11 @@ type ItemDrop struct {
 }
 
 // DataTablesParsed signals that the datatables were parsed.
-// You can use the Parser.SendTableParser() after this event to register update notification on entities & properties.
+// You can use the parser.SendTableParser() after this event to register update notification on entities & properties.
```
markus-wa: Ooops, this should reference `Parser.ServerClasses()` - my bad, either way, upper-case

```suggestion
// You can use the Parser.SendTableParser() after this event to register update notification on entities & properties.
```

examples/net-messages/README.md
```diff
@@ -49,9 +50,9 @@ cfg.AdditionalNetMessageCreators = map[int]dem.NetMessageCreator{
...
 
-When using `Parser.ParseToEnd()` net-messages and events are dispatched asynchronously. To get around this you can use `Parser.ParseNextFrame()` instead.
+When using `parser.ParseToEnd()` net-messages and events are dispatched asynchronously. To get around this you can use `parser.ParseNextFrame()` instead.
```
markus-wa:
```suggestion
When using `Parser.ParseToEnd()` net-messages and events are dispatched asynchronously. To get around this you can use `Parser.ParseNextFrame()` instead.
```

events/events.go
```diff
@@ -392,11 +392,11 @@ type ItemDrop struct {
...
 type DataTablesParsed struct{}
 
 // StringTableCreated signals that a string table was created via net message.
-// Can be useful for figuring out when player-info is available via Parser.GameState().[Playing]Participants().
+// Can be useful for figuring out when player-info is available via parser.GameState().[Playing]Participants().
```
markus-wa: upper-case

```suggestion
// Can be useful for figuring out when player-info is available via Parser.GameState().[Playing]Participants().
```

parser.go
```diff
@@ -6,17 +6,16 @@ import (
...
 )
 
 /*
 Parser can parse a CS:GO demo.
-Creating a Parser is done via NewParser().
+Creating a parser is done via NewParser().
```
markus-wa:
```suggestion
Creating a Parser is done via NewParser().
```

parser.go
```diff
@@ -220,19 +240,19 @@ type ParserConfig struct {
...
 }
 
-// DefaultParserConfig is the default Parser configuration used by NewParser().
+// DefaultParserConfig is the default parser configuration used by NewParser().
 var DefaultParserConfig = ParserConfig{
 	MsgQueueBufferSize: -1,
 }
 
-// NewParserWithConfig returns a new Parser with a custom configuration.
+// NewParserWithConfig returns a new parser with a custom configuration.
```
markus-wa: upper-case

```suggestion
// NewParserWithConfig returns a new Parser with a custom configuration.
```

parsing.go
```diff
@@ -115,22 +115,22 @@ func recoverFromUnexpectedEOF(r interface{}) error {
...
+func (p *parser) Cancel() {
 	p.cancelChan <- struct{}{}
 }
 
 /*
 ParseNextFrame attempts to parse the next frame / demo-tick (not ingame tick).
 
 Returns true unless the demo command 'stop' or an error was encountered.
-Returns an error if the header hasn't been parsed yet - see Parser.ParseHeader().
+Returns an error if the header hasn't been parsed yet - see parser.ParseHeader().
```
markus-wa: upper-case

```suggestion
Returns an error if the header hasn't been parsed yet - see Parser.ParseHeader().
```

micvbang:
Sure thing! I must admit that most of those changes are there because I simply ran a search-and-replace :sweat_smile: 

Here's a gist with `parserMock` as I've called it: https://gist.github.com/micvbang/dd985c4d6a53755a400a9c20ba870b74

It looks it we need to find a way to mock GameState as well -- because its internal fields are used throughout the package, it is quite difficult to actually mock it. One way to keep using these internal fields and allow mocking, would be to make public setters for them. But I guess that can become a quite "dangerous" API to expose. If we decide to do this, then it at least should be _extremely obvious_ from the naming that those methods are only supposed to be used for testing.

markus-wa:
I think to have everything mockable everything needs to be an interface. This would include `GameState`, `Participants`, `sendtables.{ServerClass,Entity, ...}` etc.

What's your opinion?

micvbang:
Yeah, you are right that those things all need to be hidden behind interfaces to allow for _everything_ to be mockable.

I haven't done a *lot* of regression testing yet, but from what I've done so far (from a user perspective, controlling which events come out, and in which order), I haven't really been interested in mocking `sendtables`. Right now, I feel that `sendtables` is more useful to mock for internal testing of the library itself :slightly_smiling_face: 

Yep, the events and structs in `common` and `events` are excellent as they are. All of their fields are public and easily modifiable! (`uniqueID` being the one exception, as far as I can tell)

I also haven't actually mocked `GameState` and `Participants` yet -- I tried to do my testing with as few code changes as possible, and simply added a few public setters on `GameState`, for the fields that I needed, and then whiteboxed the logic of `GameState` and `Participants` to ensure that I got the expected result. It probably would be a lot more user friendly to actually allow mocking of both of these entities! :sweat_smile:

markus-wa:
`Parser.ServerClasses()` could be interesting to mock for some cases where the user wants to get custom data from entities. Since this is a breaking change I'd rather do it too early than too late (pre-1.0.0).

`uniqueID`  should be alright as consumers can just use the `NewXYZ()` funcs if they need that.

If I find time I'll be working a bit on this as well. I'd also want to add at least one example using either https://github.com/golang/mock or https://github.com/stretchr/testify. But I'll be moving to a new place this/next week and got a lot going on at work so that might take a bit longer than usual. 😅

parser.go
```diff
@@ -156,25 +174,28 @@ Returns a identifier with which the handler can be removed via UnregisterNetMess
 
 See also: RegisterEventHandler()
 */
-func (p *Parser) RegisterNetMessageHandler(handler interface{}) dp.HandlerIdentifier {
-	return p.msgDispatcher.RegisterHandler(handler)
+
+type HandlerIdentifier *int
```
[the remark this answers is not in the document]
micvbang: It's the same issue as discussed in #40 :slightly_smiling_face:

demopacket.go
```diff
@@ -15,14 +14,60 @@ import (
 // See also: ParserConfig.AdditionalNetMessageCreators & Parser.RegisterNetMessageHandler()
 type NetMessageCreator func() proto.Message
 
+func NewNetMessageFunc(t interface{}) (int, func() proto.Message) {
```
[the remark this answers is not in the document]
micvbang: I've used it for making it easier to create NET message handlers, e.g.

```go
cfg := dem.DefaultParserConfig

cvarID, cvarFunc := dem.NewNetMessageFunc(msg.CNETMsg_SetConVar{})
cfg.AdditionalNetMessageCreators = map[int]dem.NetMessageCreator{
	cvarID: cvarFunc,
}

p := dem.NewParserWithConfig(rd, cfg)
```

Instead of

```go
cfg := dem.DefaultParserConfig
cfg.AdditionalNetMessageCreators = map[int]dem.NetMessageCreator{
	4: func() proto.Message { return new(msg.CNETMsg_Tick) },
}

p := dem.NewParserWithConfig(f, cfg)

```
It's not a pretty implementation, and it's not something that I actually meant to include in the PR.

But I do believe that it should be a bit easier to add NET handlers :slightly_smiling_face: But that should probably be done in a different thread!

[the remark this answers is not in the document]
markus-wa: Ok, yeah we can look into this separately.

I'm not 100% sure it's necessary - but anyway, topic for another PR / Issue.

parser.go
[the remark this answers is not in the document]
markus-wa: Ahhh right, I see 🙂 - I've actually been thinking of merging the `vgo` branch any time now since it's actually quite usable with go 1.11 (when outside the `GOPATH` or with `GO111MODULE=on`).

So if you are able to switch to vgo instead maybe that would be the preferred route?

Of course this will also require some docs from my end on how to use the lib like this since not everyone is used to modules yet.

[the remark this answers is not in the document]
micvbang: Yep, moving to `vgo` definitely could be a solution! That would be fine by me :slightly_smiling_face:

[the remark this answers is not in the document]
markus-wa: I've reverted the changes accordingly.

markus-wa:
I've made some changes here. I noticed my last changes were unfortunately force-pushed away it seems(?).

Some is more upper/lower case stuff, unintentional changes and the `HandlerIdentifier` / vgo thing 🙂.

Either way, just a heads up to avoid it happening again 😅.

micvbang:
Ah so sorry if I removed any changes, that definitely wasn't intentional!

micvbang:
Ugh, the godoc issue is definitely not great. 
My best bet is to look at what some larger open source codebases are doing, and consider if replicating whatever they're doing is a viable solution

markus-wa:
Sounds like a good idea. Let me know if you have a few examples, I will try to look into it as well.

markus-wa:
A similar issue in docker: https://github.com/moby/moby/issues/30454

micvbang:
Cool! For the kubernetes links, it seems like they document the struct but not the interface methods

markus-wa:
Indeed. I think this can be a viable way to do it.

I'm also considering having the interface generated from the struct methods. This should cover all the points I mentioned above (intellisense with the interface and struct, godoc, doc with implementation and maintainability).

There could then also be a check during the CI that verifies the interface doc is up-to-date.

markus-wa:
If you wan to play around with it a bit i made a branch for interface generation https://github.com/markus-wa/demoinfocs-golang/tree/interface-generator

`go generate` can be used to generate the interfaces from the structs, see https://github.com/markus-wa/demoinfocs-golang/blob/interface-generator/parser_interface.go and https://github.com/markus-wa/demoinfocs-golang/blob/interface-generator/game_state_interface.go for examples

micvbang:
Yeah, I guess that could be a viable solution -- I'd love to see how it looks.

Do you have any particular lib in mind for doing the generation?

micvbang:
Good job! Looking forward to seeing how this works :tada:

markus-wa:
Closing in favour of #66

PR #58, opened 2018-10-22
Inferno: fix bug where 0 z-dimension gives weird convex hull

micvbang:
Hey!

I just found that the update to using `z = 0` in inferno positions (7625fa0aead) makes the convex hull algorithm yield a weird result. This PR should fix it :slightly_smiling_face: 

The following screenshots are taken with with z=0 and z=1, respectively
nade_trajectories_with_0

nade_trajectories_with_1

### DIFF
--- a/common/inferno.go
+++ b/common/inferno.go
@@ -60,7 +60,7 @@ func (inf Inferno) ConvexHull2D() *s2.Loop {
 			Vector: r3.Vector{
 				X: f.Vector.X,
 				Y: f.Vector.Y,
-				Z: 0,
+				Z: 1,
 			},
 		})
 	}
--- a/common/inferno_test.go
+++ b/common/inferno_test.go
@@ -67,9 +67,9 @@ func TestInfernoConvexHull2D(t *testing.T) {
 	}
 
 	expectedHull := s2.LoopFromPoints([]s2.Point{
-		s2.Point{Vector: r3.Vector{X: 4, Y: 7, Z: 0}},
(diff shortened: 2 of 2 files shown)


markus-wa:
So there _was_ a reason why I used 1 initially 😄

Thanks Michael!

micvbang:
Huh, that's weird. That test succeeds locally!

markus-wa:
Spooky :ghost:

I'll give it a spin at home as well so we can try to figure out what's happening

micvbang:
I even have this pre-commit hook to ensure that I no longer mess with the tests! :sweat_smile: https://gist.github.com/micvbang/4c8cb1f24cfe04d1a0dfab010eb851d8

micvbang:
Just a random thought; since a lot of my PRs have had failing tests, I guess it would make sense to recommend a pre-commit hook in the README like the one I posted above to new contributors :slightly_smiling_face:

markus-wa:
No worries, that's what Travis is here for!

Personally I don't have such a hook since it takes too long (I'm a bit of a compulsory committer).
But I think I'll add a link to your gist as optional recommendation.

Is the test now failing on your machine after the change? I assume that's why you even made the change in 9af7878?

micvbang:
No, the test still works here! :slightly_smiling_face:

I believe that the problem was that while Github was behaving (very) weirdly today, Travis didn't actually get notified of each commit that I pushed, building only the first commit that did not include the fixed test

markus-wa:
Ah, that makes sense yeah. 👍

PR #59, opened 2018-10-22
datatables: default inferno fires to IsBurning=true

micvbang:
Hey Markus!

As far as I can tell, `m_bFireIsBurning` is not triggered when it starts burning, but it only appears in the list if it actually _is_ burning. So it seems that it should be defaulted to true :slightly_smiling_face:

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -438,7 +438,7 @@ func (p *Parser) bindNewInferno(entity *st.Entity) {
 				Z: float64(entity.FindProperty("m_fireZDelta." + iStr).Value().IntVal),
 			}
 
-			fire := &common.Fire{Vector: origin.Add(offset)}
+			fire := &common.Fire{Vector: origin.Add(offset), IsBurning: true}
 			entity.BindProperty("m_bFireIsBurning."+iStr, &fire.IsBurning, st.ValTypeBoolInt)
 
 			inf.Fires = append(inf.Fires, fire)


datatables.go
```diff
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	r3 "github.com/golang/geo/r3"
-
```
markus-wa: Re-add empty line (I assume that was the IDE) 🙂

I like to keep a group separate for external deps (i.e. everything that's not this repo.

micvbang: You are correct! All of my import-statement changes are from my IDE. `gofmt` I'm assuming. Perhaps another tool.

markus-wa: Probably another tool 😄 `gofmt` looks happy https://goreportcard.com/report/github.com/markus-wa/demoinfocs-golang#gofmt

micvbang: Woot! Ha, goreportcard looks cool! Thanks for showing me that.

Well, I guess visual studio has me using some other formatter then :sweat_smile:

markus-wa: Yeah it's pretty neat, I just wish it would support all of https://github.com/alecthomas/gometalinter

Btw. I use VS-Code as well and I don't think it rearranges them like that for me, maybe it's a setting?

micvbang: Hmm. I don't think so; here's my `settings.json`: https://github.com/micvbang/dotfiles/blob/master/settings.json

I'm on Ubuntu and using go v1.11 - in case your setup differs from this, could it be that there is a tool available for me that isn't for you? Just a wild guess - I can't really see how that would make sense since all of the tools (I've seen at least) are written in go

micvbang: I just recently did a `update/install go tools` in VS code. Perhaps a new tool was added to the tool chain?

markus-wa: Hmm, I'm on a fairly new windows machine (~2 weeks) so I should be pretty up-to-date as well.

I noticed I actually have `editor.formatOnSave` disabled, but it still formats when I save for some reason.
I also tried enabling it and played around with `go.formatTool` but couldn't get the same results.

PR #62, opened 2018-11-05
events: Add TeamSideSwitch, GameHalfEnd, GamePhaseChanged

micvbang:
Hey Markus!

I just stumbled upon a datatable that I'd never seen before. It seems like this datatable holds "facts" that have to be true in order for the demo to be rendered correctly in-game. 
My guess is that some of our problems w.r.t. e.g. missing `RoundEnd` events can be solved (or mitigated) using some of these fields.

I have been looking for a reliable way to detect team side switches for a long time, and I believe that the secret was hidden in `CCSGameRulesProxy` all along!

I've tested the following with approx 50 demos. They all give the same result: both `GameHalfEnd` and `TeamSideSwitch` are dispatched correctly after 15 rounds played (and are dispatched correctly for overtime as well!)

### DIFF
--- a/common/gamerules.go
+++ b/common/gamerules.go
@@ -1,0 +1,60 @@
+package common
+
+// GamePhase represents a phase in CS:GO
+type GamePhase int
+
+// The following game rules have been found at https://github.com/pmrowla/hl2sdk-csgo/blob/master/game/shared/teamplayroundbased_gamerules.h#L37.
+// It seems that the naming used in the source engine is _not_ what is used in-game.
+// The original names of the enum fields are added as comments to each field.
+const (
+	// GamePhaseInit is the default value of the game phase
+	GamePhaseInit GamePhase = 0 // enum name: Init
+
+	// GamePhasePregame
+	GamePhasePregame GamePhase = 1 // enum name: Pregame
+
+	// GamePhaseStartGamePhase is set whenever a new game phase is started.
+	// A game phase can be the normal match, i.e. first to 16 rounds, or an overtime match,
+	// i.e. first to 4 rounds. It is set for _all_ overtimes played, i.e. for a match
+	// with 3 overtimes,  GamePhaseStartGamePhase is set 1 time for the normal
+	// match and 1 time for each overtime played, for a total of 4 times.
+	GamePhaseStartGamePhase GamePhase = 2 // enum name: StartGame
+
+	// GamePhaseTeamSideSwitch is set whenever a team side switch happened,
+	// i.e. both during normal game and overtime play.
+	GamePhaseTeamSideSwitch GamePhase = 3 // enum name: PreRound
+
+	// GamePhaseGameHalfEnded is set whenever a game phase has ended.
+	// A game phase can be the normal match, i.e. first to 16 rounds, or an overtime match,
+	// i.e. first to 4 rounds. It is set once for all overtimes played, i.e. for a match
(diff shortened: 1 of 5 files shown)


common/gamerules.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// GamePhaseStaleMate has not been observed so far
+	GamePhaseStaleMate GamePhase = 6 // enum name: StaleMate
+
+	// GamePhaseGameOver has not been observed so far
+	GamePhaseGameOver GamePhase = 7 // enum name: GameOver
+)
+
+// GamePhaseToString maps a GamePhase to a user friendly string
+var GamePhaseToString = map[GamePhase]string{
```
markus-wa: This can be private

```suggestion
var gamePhaseToString = map[GamePhase]string{
```

common/gamerules.go
```diff
@@ -0,0 +1,60 @@
...
+	GamePhaseStartGamePhase: "Start game phase",
+	GamePhaseTeamSideSwitch: "Team side switch",
+	GamePhaseGameHalfEnded:  "Game half ended",
+	GamePhaseGameEnded:      "Game ended",
+	GamePhaseStaleMate:      "StaleMate",
+	GamePhaseGameOver:       "GameOver",
+}
+
+func (r GamePhase) String() string {
+	return GamePhaseToString[r]
```
markus-wa: For above
```suggestion
	return gamePhaseToString[r]
```

events/events.go
```diff
@@ -435,3 +434,20 @@ type InfernoExpired struct {
...
+	NewGamePhase common.GamePhase
+}
+
+// TeamSideSwitch signals that teams are switching sides, i.e. swapping between T and CT.
+// TeamSideSwitch is usually dispatched in the beginning of a round, just after OnRoundStart.
+type TeamSideSwitch struct {
+}
+
+// GameHalfEnded signals that the currently ongoing game half has ended.
+// GameHalfEnded is usually dispatched in the end of a round, just before OnRoundEndOfficial.
```
markus-wa: I think this should be:

```suggestion
// GameHalfEnded is usually dispatched in the end of a round, just before RoundEndOfficial.
```

datatables.go
```diff
@@ -466,3 +466,52 @@ func (p *Parser) infernoExpired(inf *common.Inferno) {
 
 	delete(p.gameState.infernos, inf.EntityID)
 }
+
+func (p *Parser) bindGameRules() {
+	grPrefix := func(s string) string {
```
markus-wa: What's your opinion on replacing this with `gameRulesPrefix + "m_gamePhase"` (or `m_totalRoundsPlayed`, respectively)?

Might be a bit clearer to see what's going on at a glance. If you agree I'd move the dot `.` into `gameRulesPrefix` too.

micvbang: My personal preference is using a function because I feel like it makes it harder to mess up the formatting, as well as being more readable to me. But that's entirely subjective and probably just because it's the way that I'm used to seeing it. Either way works for me :slightly_smiling_face:

micvbang:
Thanks for the comments!

markus-wa:
Looks like this now has the wrong base

micvbang:
Ah, thanks. That should be fixed now :slightly_smiling_face:

markus-wa:
Thanks :raised_hands: !

I also just noticed there was a comment missing for `GameState.TotalRoundsPlayed()` and added that.

PR #64, opened 2018-11-09
events: Add MatchStartChanged

micvbang:
Hey Markus!

I just ran into another funky demo [1]. In this demo, none of the previous MatchStart events are sent. But, as I had hoped for (yay!), the game rules datatable seems to contain the "truth" about certain things, among them whether a match was started or not. This PR makes it possible to read the `m_bHasMatchStarted` value from the game rules :slightly_smiling_face: 

[1]: https://www.hltv.org/matches/2327377/tyloo-vs-big-starseries-i-league-season-6

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -497,6 +497,16 @@ func (p *Parser) bindGameRules() {
 		entity.BindProperty(grPrefix("m_totalRoundsPlayed"), &p.gameState.totalRoundsPlayed, st.ValTypeInt)
 		entity.BindProperty(grPrefix("m_bWarmupPeriod"), &p.gameState.isWarmupPeriod, st.ValTypeBoolInt)
 
+		entity.FindProperty(grPrefix("m_bHasMatchStarted")).OnUpdate(func(val st.PropertyValue) {
+			oldMatchStarted := p.gameState.isMatchStarted
+			p.gameState.isMatchStarted = val.IntVal == 1
+
+			p.eventDispatcher.Dispatch(events.MatchStartedChanged{
+				OldIsStarted: oldMatchStarted,
+				NewIsStarted: p.gameState.isMatchStarted,
+			})
+		})
+
 		// TODO: seems like this is more reliable than RoundEnd events
 		// "m_eRoundWinReason"
 
--- a/datatables.go
+++ b/datatables.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	r3 "github.com/golang/geo/r3"
+
 	common "github.com/markus-wa/demoinfocs-golang/common"
 	events "github.com/markus-wa/demoinfocs-golang/events"
 	st "github.com/markus-wa/demoinfocs-golang/sendtables"
--- a/events/events.go
+++ b/events/events.go
@@ -7,7 +7,6 @@ package events
 
 import (
 	r3 "github.com/golang/geo/r3"
-
 	common "github.com/markus-wa/demoinfocs-golang/common"
 	msg "github.com/markus-wa/demoinfocs-golang/msg"
 )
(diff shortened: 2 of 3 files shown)


events/events.go
```diff
@@ -454,3 +453,10 @@ type TeamSideSwitch struct {
 // GameHalfEnded is usually dispatched in the end of a round, just before RoundEndOfficial.
 type GameHalfEnded struct {
 }
+
+// MatchStartChanged signals that the value of data table DT_GameRulesProxy.m_bHasMatchStarted has changed
+// This can be useful for some demos where the MatchStart event is not sent.
+type MatchStartChanged struct {
```
markus-wa: Should be named `MatchStartedChanged` or just `MatchStartedChange` to match the variable/function/attribute names. I'd prefer the former in this case (in combination with my other comment about moving the dispatching down, as it will then be the correct tense).

game_state.go
```diff
@@ -92,6 +93,11 @@ func (gs GameState) IsWarmupPeriod() bool {
 	return gs.isWarmupPeriod
 }
 
+// IsGameStarted returns whether the match has started.
```
markus-wa:
```suggestion
// IsGameStarted returns whether the match has started according to CCSGameRulesProxy.
```

Never know if it's actually true with all those sketchy demos out there 😄 - to avoid questions in case it's preceived to be wrong.

datatables.go
```diff
@@ -497,6 +497,14 @@ func (p *Parser) bindGameRules() {
 		entity.BindProperty(grPrefix("m_totalRoundsPlayed"), &p.gameState.totalRoundsPlayed, st.ValTypeInt)
 		entity.BindProperty(grPrefix("m_bWarmupPeriod"), &p.gameState.isWarmupPeriod, st.ValTypeBoolInt)
 
+		entity.FindProperty(grPrefix("m_bHasMatchStarted")).OnUpdate(func(val st.PropertyValue) {
+			p.eventDispatcher.Dispatch(events.MatchStartChanged{
+				OldIsStarted: p.gameState.isMatchStarted,
+				NewIsStarted: val.IntVal == 1,
+			})
+			p.gameState.isMatchStarted = val.IntVal == 1
```
markus-wa: Move this to before the dispatching, to avoid `GameState.MatchStarted()` returning the wrong value inside an event-handler.

game_state.go
micvbang: Haha, that's a good point!

micvbang:
Yeah, I have tried to see if they could be merged. By my observations that's unfortunately not the case :cry: 

I agree that there should be some documentation for when/why these events are required.

As a side note to this documentation: at one point I was considering whether events that aren't actual CSGO events , but something we came up with ("synthetic events"), should be named to signify this in some way.

events/events.go
```diff
@@ -7,7 +7,6 @@ package events
 
 import (
 	r3 "github.com/golang/geo/r3"
-
```
markus-wa: I think I've finally found something for this :smile:.

`goreturns -local github.com/markus-wa/demoinfocs-golang events/events.go` makes it 'pretty' (in my eyes). I still don't know why it always breaks in the first place, I haven't actually configured anything like that but here's what you can do in VS Code:

`"go.formatTool": "goreturns"`
and
`"go.formatFlags": ["-local github.com/markus-wa/demoinfocs-golang"]`

Unfortunately that would mean you need different workspaces for the lib and your project - so I'm not sure if that's helpful if you don't already have that.

markus-wa:
Ok, I've merged it for now. I think the doc is fine for now but I've added it to #1 nonetheless 🙂.

wrt. synthetic events, I think it's fine as is. I'm hoping to avoid too many breaking changes, even on v1.0.0.

PR #66, opened 2018-11-11
Mockable API

markus-wa:
pinging @micvbang 

This is my counter proposal to #57.

As we've discussed there I looked into interface generation to avoid godoc and Intellisense issues.
I found that another nice side-effect of this approach is that the impact on consumers is a bit lighter as they won't need to change their variable/field types from `*Parser` to `Parser` now (though for `GameState` and `Participants` that 'issue' remains).

I also added a new package `mock` that provides basic but (hopefully) extendable mocks for the various interfaces, so not everyone needs to re-implement them unless there's a specific need.

Lastly I added an example on how to use the mocks to write unit tests: https://github.com/markus-wa/demoinfocs-golang/tree/mockable-api/examples/mocking

For the last two I took a lot of inspiration from your gist which was very helpful 😄.

It would be great if you could check if there's something I missed or if this would work for you. Thanks in advance.

### DIFF
--- a/fake/parser.go
+++ b/fake/parser.go
@@ -1,0 +1,217 @@
+// Package fake provides basic mocks for IParser, IGameState and IParticipants.
+// See examples/mocking (https://github.com/markus-wa/demoinfocs-golang/tree/master/examples/mocking).
+package fake
+
+import (
+	"time"
+
+	dp "github.com/markus-wa/godispatch"
+	mock "github.com/stretchr/testify/mock"
+
+	dem "github.com/markus-wa/demoinfocs-golang"
+	common "github.com/markus-wa/demoinfocs-golang/common"
+	st "github.com/markus-wa/demoinfocs-golang/sendtables"
+)
+
+// Parser is a mock for of demoinfocs.IParser.
+type Parser struct {
+	mock.Mock
+
+	// List of events to be dispatched by frame.
+	// ParseToEnd() / ParseNextFrame() will dispatch them accordingly.
+	// See also: MockEvents() / MockEventsFrame()
+	Events map[int][]interface{}
+
+	// List of net-messages to be dispatched by frame.
+	// ParseToEnd() / ParseNextFrame() will dispatch them accordingly.
+	// See also: MockNetMessages() / MockNetMessagesFrame()
+	NetMessages map[int][]interface{}
+
+	eventDispatcher dp.Dispatcher
+	msgDispatcher   dp.Dispatcher
+	currentFrame    int
+	mockFrame       int
+}
+
+// MockEvents adds entries to Parser.Events.
+// It increments an internal frame-index so each set of events and net-messages added
+// in subsequent calls to this or MockNetMessages() is triggered on a separate frame.
+//
+// See also: MockEventsFrame()
+func (p *Parser) MockEvents(events ...interface{}) {
+	p.MockEventsFrame(p.mockFrame, events...)
+	p.mockFrame++
+}
+
+// MockEventsFrame adds entries to Events that will be dispatched at the frame indicated by the first parameter.
+//
+// See also: MockEvents()
+func (p *Parser) MockEventsFrame(frame int, events ...interface{}) {
+	p.Events[frame] = append(p.Events[frame], events...)
+}
+
+// MockNetMessages adds entries to Parser.NetMessages.
+// It increments an internal frame-index so each set of net-messages and events added
+// in subsequent calls to this or MockEvents() is triggered on a separate frame.
+//
+// See also: MockNetMessagesFrame()
+func (p *Parser) MockNetMessages(messages ...interface{}) {
+	p.MockNetMessagesFrame(p.mockFrame, messages...)
+	p.mockFrame++
+}
+
+// MockNetMessagesFrame adds entries to NetMessages that will be dispatched at the frame indicated by the first parameter.
+//
+// See also: MockNetMessages()
+func (p *Parser) MockNetMessagesFrame(frame int, messages ...interface{}) {
+	p.NetMessages[frame] = append(p.NetMessages[frame], messages...)
+}
+
+// NewParser returns a new parser mock with pre-initialized Events and NetMessages.
+// Pre-mocks RegisterEventHandler() and RegisterNetMessageHandler().
+func NewParser() *Parser {
+	p := &Parser{
+		Events:      make(map[int][]interface{}),
+		NetMessages: make(map[int][]interface{}),
+	}
+
+	p.On("RegisterEventHandler").Return()
+	p.On("RegisterNetMessageHandler").Return()
+
+	return p
+}
+
+// ServerClasses is a mock-implementation of IParser.ServerClasses().
+//
+// Unfortunately sendtables.ServerClasses currently isn't mockable.
+func (p *Parser) ServerClasses() st.ServerClasses {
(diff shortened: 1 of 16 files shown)


mock/parser.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// UnregisterEventHandler is a mock-implementation of IParser.UnregisterEventHandler().
+func (p *Parser) UnregisterEventHandler(identifier dp.HandlerIdentifier) {
+	p.eventDispatcher.UnregisterHandler(identifier)
+}
+
+// RegisterNetMessageHandler is a mock-implementation of IParser.RegisterNetMessageHandler().
+func (p *Parser) RegisterNetMessageHandler(handler interface{}) dp.HandlerIdentifier {
+	return dp.HandlerIdentifier(p.msgDispatcher.RegisterHandler(handler))
```
golangcibot: unnecessary conversion

markus-wa: good bot

micvbang:
I'm beginning to look through it now, and just an initial thought; it does not seem like I can avoid changing the type signature in my own code using the library. I still have to change it to `dem.IParser`.

This is the case because I want to my tests to test my full pipeline, including my own wrapper of the demoinfocs-golang `Parser`. In order to be able to control the behavior of `Parser`, I have to use the interface type so that I can give it the new `fake` implementation.

This is not a bad thing at all, just thought I'd note it :slightly_smiling_face:

micvbang:
I think it would be really helpful to have a primitive that allows the user to not care about the id of events per default, but allow her to do it if it really matters for the test.

In my own mock implementation I did this by keeping track of the current event id, something like this:

```go
func (p *parserMock) addEventDispatchFrame(frame int, event interface{}) {
	if frame > p.frameFinal {
		p.frameFinal = frame
	}

	p.eventDispatchQueue[frame] = append(p.eventDispatchQueue[frame], event)
	p.addFrameWithDispatch(frame)
}

func (p *parserMock) addEventDispatch(event interface{}) int {
	p.addEventDispatchFrameCurrent++
	p.addEventDispatchFrame(p.addEventDispatchFrameCurrent, event)
	return p.addEventDispatchFrameCurrent
}
```

Here, the user can let the mock track the current frame by using `addEventDispatch`, and in cases where a certain event should happen in the same frame (perhaps multiple events in the same frame), she can use `addEventDispatchFrame`. Both of them return the frame that was added, so it's easy to mix usage of the two

micvbang:
I'll also add that the most important part of this PR, to me, is the addition of the interfaces. With them, we can implement any kind of API that we want for our mocking, and everyone doesn't really have to agree about its implementation  :slightly_smiling_face:

markus-wa:
Thanks for the feedback.

I actually intentionally didn't add helpers as they are easy to add if needed, but maybe it's a good idea - might look into it again.

What I really didn't consider was multiple events in the same frame, we should certainly cover that, maybe with `map[int][]interface{}`.

And yeah the mocks in `fake` are just there as utility, if needed it's always possible to write your own.

fake/parser_test.go
```diff
@@ -0,0 +1,114 @@
...
+	assert.Equal(t, expected, actual)
+}
+
+func TestParseNextFrameEvents(t *testing.T) {
+	p := fake.NewParser()
+	p.On("ParseNextFrame").Return(true, nil)
+	expected := []interface{}{kill(common.EqAK47), kill(common.EqScout)}
+	p.Events = map[int][]interface{}{
+		0: expected,
+		1: []interface{}{kill(common.EqAUG)}, // Kill on second frame that shouldn't be dispatched
```
golangcibot: File is not `gofmt`-ed with `-s`

fake/parser_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestParseToEndEvents(t *testing.T) {
+	p := fake.NewParser()
+	p.On("ParseToEnd").Return(nil)
+	p.Events = map[int][]interface{}{
+		0: []interface{}{kill(common.EqAK47), kill(common.EqScout)},
+		1: []interface{}{kill(common.EqAUG)},
+	}
+
+	var expected []interface{}
```
golangcibot: Consider preallocating `expected`

fake/parser_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func TestParseToEndNetMessages(t *testing.T) {
+	p := fake.NewParser()
+	p.On("ParseToEnd").Return(nil)
+	p.NetMessages = map[int][]interface{}{
+		0: []interface{}{cmdKey(1, 2, 3), cmdKey(100, 255, 8)},
+		1: []interface{}{msg.CSVCMsg_Menu{DialogType: 1, MenuKeyValues: []byte{1, 55, 99}}},
+	}
+
+	var expected []interface{}
```
golangcibot: Consider preallocating `expected`

fake/parser.go
```diff
@@ -0,0 +1,200 @@
...
+
+	eventDispatcher dp.Dispatcher
+	msgDispatcher   dp.Dispatcher
+	currentFrame    int
+	mockFrame       int
+}
+
+// MockEvents adds entries to Parser.Events.
+// It increments an internal frame-index so each set of events and net-messages added
+// in subsequent calls to this or MockNetMessages() is triggered on a seperate frame.
```
golangcibot: `seperate` is a misspelling of `separate`

fake/parser.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// See also: MockEventsFrame()
+func (p *Parser) MockEvents(events ...interface{}) {
+	p.MockEventsFrame(p.mockFrame, events...)
+	p.mockFrame++
+}
+
+// MockEventsFrame adds entries to Events that will be dispatched at the frame indicated by the first paramter.
```
golangcibot: `paramter` is a misspelling of `parameter`

fake/parser.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// See also: MockEvent()
+func (p *Parser) MockEventsFrame(frame int, events ...interface{}) {
+	p.Events[frame] = append(p.Events[frame], events...)
+}
+
+// MockNetMessages adds entries to Parser.NetMessages.
+// It increments an internal frame-index so each set of net-messages and events added
+// in subsequent calls to this or MockEvents() is triggered on a seperate frame.
```
golangcibot: `seperate` is a misspelling of `separate`

fake/parser.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// See also: MockNetMessagesFrame()
+func (p *Parser) MockNetMessages(messages ...interface{}) {
+	p.MockNetMessagesFrame(p.mockFrame, messages...)
+	p.mockFrame++
+}
+
+// MockNetMessagesFrame adds entries to NetMessages that will be dispatched at the frame indicated by the first paramter.
```
golangcibot: `paramter` is a misspelling of `parameter`

markus-wa:
I fixed up the issue with multiple events on the same frame and I've also added `MockEvents(...interface{})` and `MockEventsFrame(int, ...interface{})` as per your recommendation. :slightly_smiling_face:

markus-wa:
Shall we merge this? @micvbang

micvbang:
I believe so!

markus-wa:
Done

Issue #71, opened 2019-01-07
events: Is player blinded during Kill event?


quancore:
**Higher goal**
It would be wonderful to know whether a player has been killed in a blinded situation or not. Therefore, I can update the blinded kill features.

**Describe the solution you'd like**
Probably, there should be an addition in event of Kill like:
```go
type Kill struct {
    Weapon            *common.Equipment
    Victim            *common.Player
    Killer            *common.Player
    Assister          *common.Player
    PenetratedObjects int
    IsHeadshot        bool
    IsBlinded         bool
}
```

markus-wa:
Maybe you could check `if Victim.FlashDuration > 0`. Does that work?

markus-wa:
I just discovered that `Victim.FlashDuration` doesn't always seem to work :/

Most times it's just `0` for me. Have you noticed this issue as well?

markus-wa:
I've added a helper function `Player.IsBlinded()`

quancore:
I did not test extensively for a moment, I will give feedback when done.

PR #77, opened 2019-01-11
events: add Attacker & FlashDuration() to PlayerFlashed event

### LINKED ISSUE #70: events: PlayerFlashed lack of info (Attacker / FlashDuration)
**Higher goal**
I need to know who is flashed, who did it and the duration of being blind. It is an important feature for my machine learning model.

**Describe the solution you'd like**
Probably, there should be alteration for event of PlayerFlashed like:
```go
type PlayerFlashed struct {
    Attacker *common.Player
    Defender *common.Player
    duration int
}
```

markus-wa:
Changes to `PlayerFlashed`:
- new attribute `Attacker`
- new utility function `FlashDuration()` - returns `Player.FlashDurationTime()`

Also fixes the issue with `PlayerFlashed.Player.FlashDuration` being 0 in
most cases by waiting with dispatching the events until the end of the tick.

Fixes #70 & #71

@quancore please verify

### DIFF
--- a/events/events_test.go
+++ b/events/events_test.go
@@ -1,0 +1,21 @@
+package events
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/markus-wa/demoinfocs-golang/common"
+)
+
+func TestPlayerFlashed_FlashDuration(t *testing.T) {
+	p := common.NewPlayer()
+	e := PlayerFlashed{Player: p}
+
+	assert.Equal(t, time.Duration(0), e.FlashDuration())
+
+	p.FlashDuration = 2.3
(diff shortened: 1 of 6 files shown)


events/events_test.go
```diff
@@ -0,0 +1,21 @@
+package events
+
+import (
+	"testing"
+	"time"
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

quancore:
It would be wonderful to add FlashDuration as a field to PlayerFlashed event because you are waiting at the end of the event to dispatch PlayerFlashed event so that flash duration will be available and we can return this duration in the struct as well.

PR #78, opened 2019-01-11
common: add function TeamState.Team()

### LINKED ISSUE #76: team state and team
**Higher goal**
Getting complete team information.

**Describe the solution you'd like**
The problem is that some events return team state which has nothing about the side of the team which is expected. However, there is no connection between team states and team sides (Team). For example, if I trigger scoreUpdated event, it returns me teamstate which I have no idea whether it is t or ct team. Fast solution would be adding a method in gamestate to retrieve team side by using team id.

markus-wa:
Useful for when we have a TeamState but don't know which team's it is
from context.

Fixes #76 

@quancore: does this address the issue as expected?

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -105,6 +105,8 @@ func (b Bomb) Position() r3.Vector {
 
 // TeamState contains a team's ID, score, clan name & country flag.
 type TeamState struct {
+	team Team
+
 	// ID stays the same even after switching sides.
 	ID int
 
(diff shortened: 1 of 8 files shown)


quancore:
@markus-wa  https://github.com/markus-wa/demoinfocs-golang/issues/76#issuecomment-453930500

markus-wa:
@quancore I've added your suggestions.

Do you still see some possible improvements or have any other comments?

quancore:
it seems okay.

markus-wa:
Thanks, merged

PR #87, opened 2019-01-16
common: fix regression with Inferno convex hulls

### LINKED ISSUE #85: Inferno.ConvexHull2D() returns unordered positions
**Describe the bug**
Regression introduced by 75e489d06268ee1cdb99d3829727780ad6f55c37

The result of `Inferno.ConvexHull2D()` should be ordered clockwise or counter-clockwise so that areas can be drawn easily.

This isn't the case: https://imgur.com/a/CH3tKEs

**To Reproduce**
see `examples/nade-trajectories`

**Expected behavior**
Nicely drawn areas.

**Library version**
75e489d06268ee1cdb99d3829727780ad6f55c37

**Additional context**
Discovered by @micvbang

markus-wa:
`examples/nade-trajectories` broke because the result of `ConvexHull2D()` wasn't ordered counter-clock wise anymore

- `ConvexHull2D()`: return clockwise ordered `r2.Point` slice
- `ConvexHull3D()`: return `quickhull.ConvexHull`

fixes #85

@micvbang can you verify this works for you as well?
One thing to note is that as it no longer returns a `s2.Loop` the start position is only there once, before the start and end of the loop was the same position. Depending on how your drawing code works you might need to change something.

### DIFF
--- a/common/inferno_test.go
+++ b/common/inferno_test.go
@@ -104,5 +104,5 @@ func TestInfernoConvexHull3D(t *testing.T) {
 		{X: 4, Y: 4, Z: 12},
 	}
 
-	assert.ElementsMatch(t, expectedHull, inf.ConvexHull3D(), "ConvexHull3D should contain all fire locations")
+	assert.ElementsMatch(t, expectedHull, inf.ConvexHull3D().Vertices, "ConvexHull3D should contain all fire locations")
 }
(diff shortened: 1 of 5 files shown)


common/inferno_test.go
```diff
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"github.com/golang/geo/r2"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #93, opened 2019-02-12
parsing: handle EOF and queue syncing

micvbang:


parsing.go
```diff
@@ -70,6 +70,12 @@ func (p *Parser) ParseHeader() (common.DemoHeader, error) {
 // See also: ParseNextFrame() for other possible errors.
 func (p *Parser) ParseToEnd() (err error) {
 	defer func() {
+		// Make sure all the messages of the demo are handled
+		p.msgDispatcher.SyncAllQueues()
+
+		// Close msgQueue
+		close(p.msgQueue)
```
markus-wa: Ah, nice. this could fix a potential memory leak as well.

Previously this goroutine was probably stuck indefinitely when an error occurred: https://github.com/markus-wa/godispatch/blob/master/dispatch.go#L83

👍

markus-wa: oops, I  guess the channel needs to be checked for nil before closing: https://travis-ci.org/markus-wa/demoinfocs-golang/jobs/492232647#L599

micvbang: Sweet!

markus-wa:
👍

PR #94, opened 2019-03-01
Parser: Refactor Parser.gameState.playersByUserID population

markus-wa:
Instead of checking after each tick whether there's new players to be added to `Parser.gameState.playersByUserID` we can (hopefully) do this in `Parser.bindNewPlayer()` instead.

This seems to avoid **some** game-events with `nil` values for players.
It also prevents an issue where `Participants.ByEntityID()` got out of sync and contained pointers to old players.

Also, as a bonus, it improves performance by about **5-10% (!)** according to my benchmarks.

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -179,26 +179,42 @@ func (p *Parser) bindPlayers() {
 }
 
 func (p *Parser) bindNewPlayer(playerEntity *st.Entity) {
-	var pl *common.Player
-	playerIndex := playerEntity.ID()
-	if p.gameState.playersByEntityID[playerIndex] != nil {
-		pl = p.gameState.playersByEntityID[playerIndex]
-	} else {
-		pl = common.NewPlayer()
-		p.gameState.playersByEntityID[playerIndex] = pl
-		pl.SteamID = -1
-		pl.Name = "unconnected"
-	}
+	entityID := playerEntity.ID()
+	rp := p.rawPlayers[entityID-1]
 
-	pl.EntityID = playerEntity.ID()
+	isNew := false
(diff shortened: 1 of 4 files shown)


markus-wa:
@micvbang @quancore I'd appreciate it if you guys could try this out first as I'm not sure if it maybe breaks more things than it fixes.

micvbang:
I will look into this, but I unfortunately don't have time until later in the week :)

quancore:
I am getting still nil valued events and I have tested for one demo file and nothing improved for me. Moreover, with this branch, all player has been connected with team unassigned (on master version, real players are already assigned to correct teams while they are connected which is expected).

markus-wa:
@quancore, why would it be expected that they are already assigned? Many servers let the players choose their own team. Anyway, if it's not improving anything I suppose this is of little use.

markus-wa:
And no problem @micvbang, whenerver you have time

markus-wa:
> Moreover, with this branch, all player has been connected with team unassigned (on master version, real players are already assigned to correct teams while they are connected which is expected).

@quancore this should now behave as it does in master, thought that doesn't necessarily mean players always have teams assigned on connect for every demo

markus-wa:
Adjusted according to input from @quancore (see #99) - we're now updating properties a bit sooner which is quite nice actually.

While it doesn't solve the issue discussed in https://github.com/markus-wa/demoinfocs-golang/pull/99#issuecomment-471632364 (player `z0re1` example), the behaviour should at least be like it used to be now.

The two PRs are already big enough (especially comment wise) so I think we'll solve that problem at a later point.

PR #96, opened 2019-03-02
common: fix Player.IsBlinded() inaccuracies

markus-wa:
This should fix ~~any~~ most issues with `Player.IsBlinded()`.
It may still return wrong values if the demo's header is lacking the tickrate.

It also adds another utility function `Player.FlashDurationTimeRemaining()` which returns the remaining duration of the flash. This may be useful as the player may be able to partially see if the remaining time is fairly low.

### DIFF
--- a/common/player.go
+++ b/common/player.go
@@ -49,9 +49,9 @@ func (p *Player) IsAlive() bool {
 }
 
 // IsBlinded returns true if the player is currently flashed.
-// This is more accurate than 'FlashDuration != 0' as it takes into account the FlashTick and GameState.IngameTick().
+// This is more accurate than 'FlashDuration != 0' as it also takes into account FlashTick, DemoHeader.TickRate() and GameState.IngameTick().
 func (p *Player) IsBlinded() bool {
-	return p.FlashDuration > float32(p.ingameTickProvider()-p.FlashTick)/float32(p.tickRate)
+	return p.FlashDurationTimeRemaining() > 0
 }
 
 // FlashDurationTime returns the duration of the blinding effect as time.Duration instead of float32 in seconds.
@@ -60,7 +60,28 @@ func (p *Player) FlashDurationTime() time.Duration {
 	if !p.IsBlinded() {
 		return time.Duration(0)
 	}
+	return p.flashDurationTimeFull()
+}
+
+func (p *Player) flashDurationTimeFull() time.Duration {
 	return time.Duration(float32(time.Second) * p.FlashDuration)
+}
+
+// FlashDurationTimeRemaining returns the remaining duration of the blinding effect (or 0 if the player is not currently blinded).
+// It takes into consideration FlashDuration, FlashTick, DemoHeader.TickRate() and GameState.IngameTick().
+func (p *Player) FlashDurationTimeRemaining() time.Duration {
+	// In case the demo header is broken
+	// TODO: read tickRate from CVARs as fallback
+	if p.tickRate == 0 {
+		return time.Duration(p.FlashDuration) * time.Second
+	}
+
+	timeSinceFlash := time.Duration(float64(p.ingameTickProvider()-p.FlashTick) / p.tickRate * float64(time.Second))
+	remaining := p.flashDurationTimeFull() - timeSinceFlash
+	if remaining < 0 {
+		return 0
+	}
+	return remaining
 }
 
 /*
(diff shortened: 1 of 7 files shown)


markus-wa:
@quancore could you verify if this fixes your false-positive flash-assists. Note that it does not attempt to fix anything else related to flashes, only false-positive assists.

quancore:
I have checked and it seems that it solves false positive blinded situations. However, the false negative blinded situation still exists and cause wrong stats.

markus-wa:
Good to hear.

Again, if you want me to solve false negatives you'll need to provide me a specific demo + tick where there should be a flash assist, otherwise there's nothing I can do.

markus-wa:
Just an update: I'm not quite happy with this yet.

First off, what we're doing here is sort of hacky. Second (the real deal breaker), it doesn't always work as we've seen - it breaks when the tick rate is missing from the demo header.

@micvbang have you got an idea how we could fix this nicely?

common/player.go
```diff
@@ -30,30 +30,60 @@ type Player struct {
...
+	FlashTick                   int     // In-game tick at which the player was last flashed
 	Team                        Team
 	TeamState                   *TeamState // When keeping the reference make sure you notice when the player changes teams
 	IsBot                       bool
 	IsDucking                   bool
 	IsDefusing                  bool
 	HasDefuseKit                bool
 	HasHelmet                   bool
+
+	tickRate float64 // the in-game tick rate, used for IsBlinded()
```
golangcibot: File is not `goimports`-ed (from `goimports`)

common/player_test.go
```diff
@@ -51,20 +52,73 @@ func TestPlayerAlive(t *testing.T) {
...
+	pl.FlashDuration = 1
+	pl.FlashTick = 128
+	*pl.ingameTick = 128 * 4
+	assert.Equal(t, time.Duration(0), pl.FlashDurationTimeRemaining())
+}
 
-	assert.Equal(t, 2300*time.Millisecond, p.FlashDurationTime())
+func newPlayer() *Player {
+	tick := 0
+	return NewPlayer(128, &tick)
```
golangcibot: cannot use &tick (value of type *int) as IngameTickProvider value in argument to NewPlayer (from `typecheck`)

common/player_test.go
```diff
@@ -51,20 +52,73 @@ func TestPlayerAlive(t *testing.T) {
...
+	*pl.ingameTick = 128 * 4
+	assert.Equal(t, time.Duration(0), pl.FlashDurationTimeRemaining())
+}
+
+func TestPlayer_FlashDurationTimeRemaining_FlashDuration_Over(t *testing.T) {
+	pl := newPlayer()
+
+	pl.FlashDuration = 1
+	pl.FlashTick = 128
+	*pl.ingameTick = 128 * 4
```
golangcibot: invalid operation: `pl` (variable of type `*Player`) has no field or method `ingameTick` (from `typecheck`)

events/events_test.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 func TestPlayerFlashed_FlashDuration(t *testing.T) {
-	p := common.NewPlayer()
+	tick := 0
+	p := common.NewPlayer(128, &tick)
```
golangcibot: cannot use &tick (value of type *int) as github.com/markus-wa/demoinfocs-golang/common.IngameTickProvider value in argument to common.NewPlayer (from `typecheck`)

fake/parser_test.go
```diff
@@ -51,10 +51,11 @@ func TestParseNextFrameEvents(t *testing.T) {
 
 func kill(wepType common.EquipmentElement) events.Kill {
 	wep := common.NewEquipment(wepType)
+	tick := 0
 	return events.Kill{
-		Killer: common.NewPlayer(),
+		Killer: common.NewPlayer(128, &tick),
```
golangcibot: cannot use &tick (value of type *int) as github.com/markus-wa/demoinfocs-golang/common.IngameTickProvider value in argument to common.NewPlayer (from `typecheck`)

markus-wa:
I've changed it so that if `DemoHeader.TickRate() == 0` then a fallback to `FlashDuration` is used (behaves as it is right now).

Hopefully in the future we'll find a way to get the tick rate for the demos where it's missing now so we can ditch the fallback.

@quancore - let me know if this does the trick

quancore:
> I've changed it so that if `DemoHeader.TickRate() == 0` then a fallback to `FlashDuration` is used (behaves as it is right now).
> 
> Hopefully in the future we'll find a way to get the tick rate for the demos where it's missing now so we can ditch the fallback.
> 
> @quancore - let me know if this does the trick

Right now, I have no demo file with missing tick rate to check it. If you are okay for you, it is okay.

PR #104, opened 2019-03-25
metadata: refactoring

o40:


markus-wa:
Can you restore the `MapXXX` vars? Otherwise it's a breaking change unfortunately.

If v1 wasn't out I'd prefer your version as well though :slightly_smiling_face:.

I think it might also be better to make `MakeMap` private, in case we want to add new fields to the `Map` struct later on (to avoid future breaking changes). What do you think?

o40:
I agree. Creating new pull request.

PR #105, opened 2019-03-29
metadata: refactoring

o40:
Updated version of the refactoring where the public variables are not dropped.


markus-wa:
LGTM, thanks!

PS, you coukd have reused the old PR and either force pushed or added a fix commit ( I'd have squashed it in that case). That way we keep some history and discussions in one place 🙂

PR #107, opened 2019-03-31
Carrying weapons IDs of Player

ArlindoTorres:
Useful functions specially when working with cs-demo-minifier for extracting all weapon id's of a player at each tick for example. Provides better performance and smaller generated file when minifying a demo.

### DIFF
--- a/common/player.go
+++ b/common/player.go
@@ -105,11 +105,25 @@ func (p *Player) ActiveWeapon() *Equipment {
 	return p.RawWeapons[p.ActiveWeaponID]
 }
 
+// ActiveWeaponID returns the currently active / equipped weapon ID of the player.
+func (p *Player) ActiveWeaponID() int {
+	return int(p.RawWeapons[p.ActiveWeaponID].Weapon)
+}
+
 // Weapons returns all weapons in the player's possession.
 func (p *Player) Weapons() []*Equipment {
 	res := make([]*Equipment, 0, len(p.RawWeapons))
 	for _, w := range p.RawWeapons {
 		res = append(res, w)
+	}
+	return res
+}
+
+// WeaponIDs returns all weapon IDs in the player's possession.
+func (p *Player) WeaponIDs() []int {
+	res := make([]int, 0, len(p.RawWeapons))
+	for _, w := range p.RawWeapons {
+		res = append(res, int(w.Weapon))
 	}
 	return res
 }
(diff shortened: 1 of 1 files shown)


common/player.go
```diff
@@ -105,6 +105,11 @@ func (p *Player) ActiveWeapon() *Equipment {
 	return p.RawWeapons[p.ActiveWeaponID]
 }
 
+// ActiveWeaponID returns the currently active / equipped weapon ID of the player.
+func (p *Player) ActiveWeaponID() int {
```
golangcibot: field and method with the same name ActiveWeaponID (from `typecheck`)

markus-wa:
Thanks for the PR.

Calling the function `WeaponIDs` might be a bit misleading, as it's not the same thing as for instance `ActiveWeaponID`. `ActiveWeaponID` is the key of the currently active weapon in `RawWeapons` while what's being returned here is the `EquipmentElement`.
A better name might be `WeaponTypes()`, though I'm not super happy about that one either since it doesn't conform with the libs naming scheme - and `WeaponEquipmentElements()` would sound wrong and irritating.

Aside from that I'm not sure if it's necessary to do this in the lib, maybe it should be done on the consumer side (i.e. in `cs-demo-minifier`)? Since it's really just a `map()` function (it's a bit unfortunate go doesn't support them natively in a convenient way).

I'd suggest putting this into `cs-demo-minifier`:
```go
func myFunc(p *Player) {
    wepTypes := weaponTypes(p.Weapons())
    // TODO: do something interesting with wepTypes
}

func weaponTypes(weps []*common.Equipment) []int {
    types := make([]string, len(weps))
    for i, w := range weps{
        types[i] = w.Weapon
    }
    return types
}
```

What do you think?

markus-wa:
Regarding performance: cs-demo-minifier shouldn't do something every tick, only on the snapshot ticks. 

If you think there is actually a non-negligible performance issue / bottleneck due to the additional slice being created in my suggestion we should find a better solution, like dispatching events here:
https://sourcegraph.com/github.com/markus-wa/demoinfocs-golang@8f7845ce30aca0c1fdb406734ec7a637634a72a6/-/blob/datatables.go#L279
and then only doing updates in the minifier after these events occur

ArlindoTorres:
When I talked about doing each tick I really meant each snapshot tick.
But yes, you're right I'll do it on minifier instead.

Issue #108, opened 2019-04-06
divide by zero in demoinfocs-golang/common.DemoHeader.FrameTime


marksamman:
**Describe the bug**
Running the code below on the demo file linked results in a panic (runtime error) because it's dividing by zero in common.DemoHeader.FrameTime.

**To Reproduce**
https://www.dropbox.com/s/z0xjroygkdec6l6/742301.dem?dl=0

Code:
```go
		f, err := os.Open("742301.dem")
		if err != nil {
			panic(err)
		}
		defer f.Close()

		p := dem.NewParser(f)
		p.RegisterEventHandler(func(e events.SayText2) {
			 p.CurrentTime()
		})
		p.ParseToEnd()
```

**Library version**
v1.1.0

markus-wa:
Yikes.

This looks like another case of info being missing in the demo header of some demos.

Unfortunately I don't have access to a PC the next 3 weeks so it'll have to wait.

The fix will probably be to return 0 if PlaybackTime is 0

markus-wa:
If you could create a PR with the fix I'd be able to merge it though 🙂

marksamman:
Submitted a pull request: #109. Any chance you could tag a release after merging? 🙂

markus-wa:
> Any chance you could tag a release after merging?

Yes sir

markus-wa:
Thanks for the fix, I've tagged it with v1.1.1 ✌🏻

PR #109, opened 2019-04-07
Fix divide by zero in common.DemoHeader.FrameTime

marksamman:
See #108

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -41,6 +41,8 @@ func (h DemoHeader) FrameRate() float64 {
 }
 
 // FrameTime returns the time a frame / demo-tick takes in seconds.
+//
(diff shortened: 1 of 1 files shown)


common/common.go
```diff
@@ -42,6 +42,9 @@ func (h DemoHeader) FrameRate() float64 {
 
 // FrameTime returns the time a frame / demo-tick takes in seconds.
```
markus-wa:
```suggestion
// FrameTime returns the time a frame / demo-tick takes in seconds.
//
// Returns 0 if PlaybackTime or PlaybackFrames are 0 (corrupt demo headers).
```

markus-wa:
Awesome, thanks.

Can you apply my suggestion? Says I'm not allowed to.

PR #133, opened 2019-07-27
Equipment.AmmoReserve

xXxhaxmeister42xXx:
Added two functions to retrieve all gamephase- and weapon names to allow direct export along with match data.

Equipment.AmmoReserve has been left uninitialized at 0 so far, so i thought i might add that as well.

Not sure if i missed something aside from properly recreating the Interface, reserve ammo update seemed pretty straightforward.

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -419,6 +419,9 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement) 
 	entity.FindPropertyI("m_iClip1").OnUpdate(func(val st.PropertyValue) {
 		eq.AmmoInMagazine = val.IntVal - 1
 	})
+	entity.FindPropertyI("m_iPrimaryReserveAmmoCount").OnUpdate(func(val st.PropertyValue) {
+		eq.AmmoReserve = val.IntVal
+	})
 
 	// Only weapons with scopes have m_zoomLevel property
 	if zoomLvlProp := entity.FindPropertyI("m_zoomLevel"); zoomLvlProp != nil {
--- a/datatables.go
+++ b/datatables.go
@@ -419,9 +419,10 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement) 
 	entity.FindPropertyI("m_iClip1").OnUpdate(func(val st.PropertyValue) {
 		eq.AmmoInMagazine = val.IntVal - 1
 	})
-	entity.FindPropertyI("m_iPrimaryReserveAmmoCount").OnUpdate(func(val st.PropertyValue) {
-		eq.AmmoReserve = val.IntVal
-	})
+	// Some weapons in some demos might be missing this property
+	if reserveAmmoProp := entity.FindPropertyI("m_iPrimaryReserveAmmoCount"); reserveAmmoProp != nil {
+		reserveAmmoProp.Bind(&eq.AmmoReserve, st.ValTypeInt)
+	}
 
 	// Only weapons with scopes have m_zoomLevel property
 	if zoomLvlProp := entity.FindPropertyI("m_zoomLevel"); zoomLvlProp != nil {


markus-wa:
Thanks for the PR!

Could we seperate the helpers into another PR?

What are you using the `EqElementToName` and `GamePhaseToString` maps for? I don't think we should export the two but maybe I'm missing something.

Also Travis isn't quite happy yet, you need to install https://github.com/vburenin/ifacemaker in your PATH and use `go generate` to create the interfaces 😄. I saw you tried to do it manually but it checks if it is identical to the generated one 🙂.
I'll have to document this in the README!

xXxhaxmeister42xXx:
Well, for some haxmeisters that are new to Go that could be helpful, i guess ¯\_(ツ)_/¯

Seems travis still isnt happy. I'm not quite sure how to split pull requests? Revert?

I am working on a match export/conversion script for standalone simple renderers, exporting to JSON or custom formats. To ensure (weapon) name/ID consistency ElementToName, GamePhases etc are added as well.
Not a big deal to add export functions in my local copies before compiling, but a few extra functions wouldnt hurt either imo.

markus-wa:
You dropped a \ - ¯\\\_(ツ)\_/¯ - `¯\\\_(ツ)\_/¯`.

You have a couple of options, here's a few:

A (rebase, suggested):
```
git checkout -b ammo-reserve
git rebase -i ff6df95^
# now vi opens. delete all commits you don't want in this PR
git push -u origin ammo-reserve

git checkout master && git checkout -b helper-functions
git checkout helper-functions
git rebase -i ff6df95^
# delete all commits you don't want in the helper functions PR
git push -u origin helper-functions
```
Then change the base-branch in Github from ` xXxhaxmeister42xXx:master` to ` xXxhaxmeister42xXx:ammo-reserve`


B (cherry-pick):
```
git remote add upstream https://github.com/markus-wa/demoinfocs-golang.git
git fetch upstream
git checkout -b ammo-reserve
git reset --hard upstream/master # reset new branch to the current master of this repo
git cherry-pick ff6df95 # take the ammo-reserve commit and add it to the new branch
git push -u origin ammo-reserve

git checkout master && git checkout -b helper-functions
git reset --hard upstream/master
git cherry-pick <space-seperated list of commits for helper-functions>
git push -u origin helper-functions
```
Then change the base-branch in Github from ` xXxhaxmeister42xXx:master` to ` xXxhaxmeister42xXx:ammo-reserve`

C (backup + hard reset):
```
git checkout -b helper-functions # save other commits in a new branch
git rebase -i ff6df95^
# delete the line with ff6df95, this makes the helper-functions branch a bit cleaner
git push -u origin helper-functions

git checkout master
git reset --hard ff6df95
git push -f # force push, removes new commits
```

Regarding the name export functions. I see your point now. In that case, can we name the functions `EquipmentElementNames` and `GamePhaseNames`.
Please also add comments to the functions according to Effective Go:

> Inside a package, any comment immediately preceding a top-level declaration serves as a doc comment for that declaration. Every exported (capitalized) name in a program should have a doc comment.

> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared. 

So the first word of the comment should be the function name. Otherwise Travis won't be happy either after fixing the current issue.

You can check out Travis's  issues by clicking on `details`, then Go 1.11 (for example).
In this case the issue is that we didn't update the mocks in `fake/game_state.go`: https://travis-ci.org/markus-wa/demoinfocs-golang/jobs/564472769#L287
You can take any of the existing mock functions as reference.

markus-wa:
Sorry for the wall of text haha 😄

xXxhaxmeister42xXx:
Wohoo, Nullpointerexception.

Not all weapons with m_iPrimaryReserveAmmoCount? Would be interesting to know which one.
I did run a few demos.

markus-wa:
I suspect it's the first demo in the regression set from 2015.

This happens with a few properties, here's how to fix it:
https://github.com/markus-wa/demoinfocs-golang/blob/master/datatables.go#L178

xXxhaxmeister42xXx:
Yeah i saw the nil check for m_zoomLevel, but since even knives etc had the ammo property i assumed it to be unneccessary.

Any reason you mix FindPropertyI(...).Property.OnUpdate with Property.Bind, instead of adding a nil check to Entity.BindProperty?

https://github.com/markus-wa/demoinfocs-golang/blob/d03a884d868fd585560885d7f8fefc38759895b9/sendtables/entity.go#L79

markus-wa:
Oh right, we should use `Bind` instead of `OnUpdate` if possible.

But as for the check in `BindProperty`: I think it's better to not handle this silently. Debugging a property not being set could otherwise get a bit nasty.
This way we notice it during the implementation and can add comments to these fields explaining that they might not always work (which we should actually do in this case, I forgot!).

markus-wa:
PS: this should get fixed once we get rid of all the property-bindings in `v2.0.0` (soonTM)

xXxhaxmeister42xXx:
<details>
<summary>The actual implementation could look like</summary>

```go
func (e *Entity) BindProperty(name string, variable interface{}, valueType PropertyValueType) 
	e.FindPropertyI(name).onUpdate(getBinder(variable, valueType))

func (e *Entity) BindPropertyIfExists(name string, variable interface{}, valueType PropertyValueType)
func (e *Entity) BindPropertyHandler(name string, handler PropertyUpdateHandler)
func (e *Entity) BindPropertyHandlerIfExists(name string, handler PropertyUpdateHandler)

func getBinder(variable interface{}, valueType PropertyValueType) PropertyUpdateHandler {
	switch valueType {
	case ValTypeInt:
		return func(val PropertyValue) {
			*(variable.(*int)) = val.IntVal
		}
	...
```

</details>

I added a check for m_iPrimaryReserveAmmoCount before binding like the other properties have.

markus-wa:
Awesome, thanks!

That `BindPropertyIfExists`, `BindPropertyHandler` etc. stuff adds a lot of noise to the API.
I'd actually rather even remove `BindProperty` altogether and just have the consumer (in this case `datatables.go`) do everything itself.

The consumer could also do things like this:
```
func bindProperty(e st.IEntity, prop string, var interface{}, type PropertyValueType) {
 ...
}

func bindPropertyIfExists(e st.IEntity, prop string, var interface{}, type PropertyValueType) {
...
}
```

This would solve the problem and also hide the helpers to the public.

PR #134, opened 2019-07-29
GameState: Allow accessing .gameState and .currentDefuser

xXxhaxmeister42xXx:
Mock functions okay like this or should GamePhase return some default phase and CurrentDefuser nil?

https://github.com/xXxhaxmeister42xXx/demoinfocs-golang/blob/c386294a57347adc0c8dc11a606049b4866b66a5/fake/game_state.go#L68-L71

https://github.com/xXxhaxmeister42xXx/demoinfocs-golang/blob/c386294a57347adc0c8dc11a606049b4866b66a5/fake/game_state.go#L83-L86


markus-wa:
I'd like to avoid exposing `currentDefuser`.
This should be handled with events in the consumer app.

see
- [`BombDefuseStart`](https://godoc.org/github.com/markus-wa/demoinfocs-golang/events#BombDefuseStart)
- [`BombDefuseAborted`](https://godoc.org/github.com/markus-wa/demoinfocs-golang/events#BombDefuseAborted)
- [`BombDefused`](https://godoc.org/github.com/markus-wa/demoinfocs-golang/events#BombDefused)

xXxhaxmeister42xXx:
Well, if you insist on keeping it disclosed i guess i can do one more revert to end the day ^^

So far i cant see a reason _not_ to expose it, unless you want to remove it. To me it looked more like a missing getter function as with gamephase.
Yes, it would not be a big deal to store data from the events and nonlinear parsing/seeking is probably not the first on your to-do-list, but the pointer to the player defusing is already there and could be used without downside or checking all players or the bomb.

markus-wa:
I don't understand why you would need to check all players or the bomb. You already have the events.
Or is there something missing/broke.

The reason not to expose it is because it makes the API less clean if we have tons of ways to so the same thing. This is already an issue today with events and getters and I'd like to avoid making it worse.

`defusingPlayer` is only relevant for such tinyparts of the demo (when someone is actually defusing) that it doesn't make sense to have a  getter on the `GameState` imo.

Maybe I'm missing something. Can you share the code of what you're trying to do?

xXxhaxmeister42xXx:
I'm not trying to do something, but my approach would be: The reference exists and therefore could be made accessible.
But nonlinear parsing set aside i cant think of a situation where it would actually be required.

markus-wa:
Ah I see, fair point.

I prefer to go the way of making as little as possible accessible but as much as needed.
You can always add new functions but you can't remove them (without breaking backwards compatibility) 🙂

PR #143, opened 2019-08-30
fix delayedEventHandlers not being updated

### LINKED ISSUE #142: ItemEquip with a nil player pointer
**Describe the bug**
In the demo linked below there is an event of item being equipped, but in the dispatched event the Player pointer appears to be `nil`. I researched it as far as I understand the library code and found that:
1) in the raw event data `userid` is 42;
2) `geh.gameState().playersByUserID[42]` is `nil`;
3) `geh.parser.rawPlayers[14].userID == 42`, so the player actually exists in the game. It is a bot btw.

Unfortunately, I cannot say what exactly is causing this behaviour. But there is several round restoring & restarting prior to the event so I assume it is somehow connected.

**To Reproduce**
https://www.hltv.org/download/demo/30352

Code:
Conditional breakpoint at game_events.go:67 (`playerByUserID` first line) with userID == 42 as the condition.

**Expected behavior**
e.Player pointing to the player with userid == 42.

**Library version**
v1.3.1

BestAwperEver:
Also delayed `player_blind` event
Should finally fix #142 :)

### DIFF
--- a/game_events.go
+++ b/game_events.go
@@ -73,7 +73,11 @@ func newGameEventHandler(parser *Parser) gameEventHandler {
 	geh := gameEventHandler{parser: parser}
 
 	delay := func(f gameEventHandlerFunc) gameEventHandlerFunc {
-		return func(data map[string]*msg.CSVCMsg_GameEventKeyT) { f(data) }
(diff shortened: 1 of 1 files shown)


markus-wa:
Cheers for the fix! ✌️

PR #178, opened 2019-12-29
WIP: bomb damage detection

### LINKED ISSUE #169: Can't detect bomb damage
**Describe the bug**
`PlayerHurt.Weapon` is unkown when a player is receiving bomb explosion damage

**To Reproduce**
At the end of round 25 of this demo, 2 players are receiving bomb damage : https://gofile.io/?c=Je2o0z

Code:
```go
parser.RegisterEventHandler(func(e events.MatchStart) {
    round := 1

    parser.RegisterEventHandler(func(e events.RoundStart) {
        round++
    })

    parser.RegisterEventHandler(func(e events.PlayerHurt) {
        if 25 == round {
            fmt.Printf("[PlayerHurt] %s receive %d damage from weapon %s\n", e.Player.Name, e.HealthDamage, e.Weapon.String())
        }
    })
})
```

Current result: 
```
demo-analyzer-api    | [PlayerHurt] [B]4ll receive 277 damage from weapon UNKNOWN
demo-analyzer-api    | [PlayerHurt] Luckso receive 7 damage from weapon UNKNOWN
demo-analyzer-api exited with code 0
```

**Expected behavior**
The weapon should be the bomb

markus-wa:
cc @xavier-rodet @BestAwperEver

I tried to fix #169 but it's more complicated than I had hoped (mostly due what @BestAwperEver mentioned [here](https://github.com/markus-wa/demoinfocs-golang/issues/169#issuecomment-569301938).

Here is my attempt, but I don't think I will end up merging it, certainly not in this state: https://github.com/markus-wa/demoinfocs-golang/compare/issue/169-bomb-damage-detection#diff-4af3cd88b0d18d12411b7348b52d476cR680

It seems to work for all demos in the regression set but there could be issues with older demos where the default value for `mp_c4time` was different, since we assume 45 seconds if it's not set.

image

### DIFF
--- a/game_events.go
+++ b/game_events.go
@@ -3,6 +3,7 @@ package demoinfocs
 import (
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/golang/geo/r3"
 	"github.com/markus-wa/go-unassert"
@@ -479,9 +480,18 @@ func (geh gameEventHandler) bombBeginPlant(data map[string]*msg.CSVCMsg_GameEven
 }
 
 func (geh gameEventHandler) bombPlanted(data map[string]*msg.CSVCMsg_GameEventKeyT) {
+	gs := geh.gameState()
+	gs.currentPlanter.IsPlanting = false
+	gs.currentPlanter = nil
+
+	c4Timer, err := strconv.Atoi(gs.conVars["mp_c4timer"])
+	if err != nil {
+		c4Timer = 45
+	}
+	gs.nextBombExplosionTime = geh.parser.CurrentTime() + time.Second*time.Duration(c4Timer)
+
 	event := events.BombPlanted{BombEvent: geh.bombEvent(data)}
-	event.Player.IsPlanting = false
-	geh.parser.gameState.currentPlanter = nil
+	event.Player.IsPlanting = false // just to be sure, 'should' be the same as above
 	geh.dispatch(event)
 }
 
@@ -492,8 +502,11 @@ func (geh gameEventHandler) bombDefused(data map[string]*msg.CSVCMsg_GameEventKe
 }
 
 func (geh gameEventHandler) bombExploded(data map[string]*msg.CSVCMsg_GameEventKeyT) {
+	gs := geh.gameState()
+	gs.currentDefuser = nil
+	gs.bombExplosions[gs.ingameTick] = true
+
 	bombEvent := geh.bombEvent(data)
-	geh.gameState().currentDefuser = nil
 	geh.dispatch(events.BombExplode{BombEvent: bombEvent})
 }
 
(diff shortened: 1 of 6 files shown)


game_events.go
```diff
@@ -479,9 +480,18 @@ func (geh gameEventHandler) bombBeginPlant(data map[string]*msg.CSVCMsg_GameEven
...
 func (geh gameEventHandler) bombPlanted(data map[string]*msg.CSVCMsg_GameEventKeyT) {
+	gs := geh.gameState()
+	gs.currentPlanter.IsPlanting = false
+	gs.currentPlanter = nil
+
+	c4Timer, err := strconv.Atoi(gs.conVars["mp_c4timer"])
+	if err != nil {
+		c4Timer = 45
+	}
+	gs.nextBombExplosionTime = geh.parser.CurrentTime() + time.Second*time.Duration(c4Timer)
```
golangcibot: assignments should only be cuddled with other assignments (from `wsl`)

game_events.go
```diff
@@ -661,6 +674,14 @@ func (geh gameEventHandler) deleteThrownGrenade(p *common.Player, wepType common
 }
 
 func (geh gameEventHandler) getEquipmentInstance(player *common.Player, wepType common.EquipmentElement) *common.Equipment {
+	gs := geh.gameState()
+	bombExplodedThisTick := gs.bombExplosions[geh.gameState().ingameTick]
+	diff := gs.nextBombExplosionTime - geh.parser.CurrentTime()
+	bombExplodedThisTick = bombExplodedThisTick || diff < 5*time.Second && diff > -5*time.Second
+	if wepType == common.EqUnknown && bombExplodedThisTick {
```
golangcibot: only one cuddle assignment allowed before if statement (from `wsl`)

BestAwperEver:
About c4 timer: from what I know it is 40 seconds in MM, majors and most cs go tournaments (though in rare cases it is 35), so in my application I use this value as a fallback. Well, in reality it is only technically a "fallback" because I has not yet encounter a demo which I could fetch the `mp_c4timer` cvar from.

If you will eventually merge this into the library, it would be nice to have it as an optional feature, not as hardcoded one, because there inevitably will be some cases where the implemented logic might be wrong.

markus-wa:
We *could* add a field to `ParserConfig` but honestly I'd rather avoid config options. Because even if it's an optional choice it should work for the case where it's enabled, which we really can't be sure of.

What I'm really hoping for is some other, bulletproof way, to detect explosions.

BestAwperEver:
Well, for now the only way I see it working is to delay _all_ events to the end of the tick, the dispatch them in some "right" order (e.g. first kill events, then bomb explosion, or may be leave them as is, just with additional info) but this time with additional data from the further events.

It is quite a goal and might be added to something like "2.0 milestone" but it has its advantages beside bomb damage determination. And, of course, it will work only as long as we can be sure that the bomb has exploded in the current tick, that is, if we have `bomb_exploded` event which theoretically might be absent in some demos (?).

markus-wa:
> delay all events to the end of the tick

I already tried that once, it didn't end very well at all 😓.

The issue is that at the end of the tick, some info that's needed for many events is completely gone. I don't remember the exact issues but there were plenty and it just lead to a lot of confusing contradictions. The only reason we do delay some events is because the opposite also happens (data is only available at the end of the tick.

The root cause is that entity update packages are sent after game events, and sometimes those updates are needed and sometimes they need to be applied after the event.

markus-wa:
Maybe we can delay things a bit less than to the end of the tick. We could collect all non-delayed game events and then dispatch them before the entity updates go through. That should preserve todays ordering while giving us the chance to detect c4 damage (if the explosion event is present, which i think it always is). Even if it's not always present its still an improvement over today.

I'm not sure if this will work but I'll give it a go some time.

markus-wa:
closing in favour of #205

PR #192, opened 2020-03-24
added Weapons() method to parser

BestAwperEver:
Comes quite handy when working, for example, with raw event data.

This time I couldn't deal with ifacemaker at all, even without any changes in the project I get `121:1: comment not terminated (and 1 more errors)`. I tried to run it from IDE/cmd/powershell/git bash/WSL but all in vain 😞 so I manually redacted the `parser_interface.go` and am hoping that I got lucky with method declaration order 😃


parser.go
```diff
@@ -157,6 +157,13 @@ func (p *Parser) Progress() float32 {
 	return float32(p.currentFrame) / float32(p.header.PlaybackFrames)
 }
 
+// Weapons returns array which is used to remember what a weapon is (p250 / cz etc.).
+//
+// Note: this array is filled with meaningful values after DataTablesParsed event is dispatched.
+func (p *Parser) Weapons() [maxEntities]common.Equipment {
```
markus-wa: I think we'll want to return a map here. I never really liked the fact we were using an array for weapons anyway.

Also it should be a copy of the underlying data, to avoid giving consumers the power to overwrite internal parser data.

Iirc this will take some work to adjust properly, last time I tried this it didn't go so well.

BestAwperEver: I understand your concerns about giving a consumer the ablity to change the data, but in this particular case I actually used the fact that in Go arrays are passed by value so this actually _is_ a copy 🙂 That's the reason I [added a test to parser_test.g](https://github.com/markus-wa/demoinfocs-golang/blob/49ed3f041d7be50abf12941c376505fa6622e9e2/parser_test.go#L69)o which ensures that we cannot change the original array.

As for returning a map, yeah, this may be better for the end-user understanding and more in line with other library methods. The reason I didn't make a map is the efficiency of copying an array using  golang internals vs making a new map and filling it. It is quite a pity in Go we cannot pass a map by constant reference like in c++ -- actually being a former c++ coder it hurts my feelings to know that every time I use something like `GameState.Participants.Playing()` a full copy of the map is created 😃 So, if you really think we should return a map, let me know and I certainly will change the behaviour. I don't really think it will strike the performance of my application since it already uses quite a lot of map-copying.

TBH though I would be happy if there were more methods like [this](https://github.com/markus-wa/demoinfocs-golang/blob/be93db4b8c9cbc60cf0942b52ab455b1f29b50bf/parser.go#L102) which do not return a copy and return a map reference instead with remark about "Intended for internal usage" or something like that, so user could use them on their own head with theoretical possibility of internal data corruption but without having to copy a map every time they need access to the data.

Could you tell me a bit more about difficulties you encountered the last time so I could make some adjustments? Right now I don't really understand whan do you mean because I've already used this fork to draw a simple 2D view of a game and didn't faced any problems with it so far.

markus-wa: I'm a bit suprised to see that this test passes. Slices are actually passed as a `SliceHeader` which contains a pointer to the underlying array, so it should be like passing by reference (https://stackoverflow.com/questions/39993688/are-golang-slices-passed-by-value).

And what I meant is to not only change the `Weapons()` function but also change the internal representaton from a slice to a map (`Parser.weapons`). I will try to look into this again - you can give it a go as well if you feel like it of course.

And yeah Go's immutability support is terrible, but considering the amount of CPU time spent on processing a demo file, the work to copy a map is so tiny that it's negligible imo. The benefits outweigh the negatives for me. What I wouldn't want to do is turn the slice into a map only inside `Weapons()`, in that case i'd prefer your current implementation

BestAwperEver: I suppose we all work so often with slices in Go that sometimes we forget that such thing as "an array" does even exist 😄 

Well I don't feel like I could break something if I replace your current array-representation with a map one myself. Don't even have to replace the tese code 😃 So I think I will add the changes to the PR so you could review them later.

game_events.go
```diff
@@ -694,7 +694,7 @@ func getPlayerWeapon(player *common.Player, wepType common.EquipmentElement) *co
 	}
 
 	wep := common.NewEquipment(wepType)
-	return &wep
+	return wep
```
golangcibot: return statements should not be cuddled if block has more than two lines (from `wsl`)

game_events.go
```diff
@@ -694,7 +694,8 @@ func getPlayerWeapon(player *common.Player, wepType common.EquipmentElement) *co
 	}
 
 	wep := common.NewEquipment(wepType)
-	return &wep
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

BestAwperEver: @markus-wa need your help here 😃 @golangcibot is doing this again

The same issue as the last time: first he forced me to add a new line and now he wants it removed just to complain again about return statement being cuddled. This is an endless cycle and I still have no idea what should be done to please him.

BestAwperEver:
I ran the regression tests locally and everything seems to be OK. Lets see how it goes with Travis.

Also added `Parser.WeaponTypeByEntityID` method since it is the primary use of the underlying map. Probably this were sufficient from the beggining without any other changes 😃 

@codeclimate found 2 duplicate code fragments in some test code I didn't touch so I think that's ok?

datatables.go
```diff
@@ -302,7 +302,15 @@ func (p *Parser) bindNewPlayer(playerEntity st.IEntity) {
...
 				cache[i2] = entityID
 
-				wep := &p.weapons[entityID]
+				wep := p.weapons[entityID]
+
+				if wep == nil {
+					// Something is clearly wrong here
+					// But since we had an empty Equipment instance here before the change
+					// from array with default elements to map, let's create a new instance.
+					wep = new(common.Equipment)
```
markus-wa: note that we won't have a `UniqueID` in this case. I know it wasn't there before either, but I'm wondering if we should add it or not. I think this could actually be a common case (or maybe it's indeed never used. unfortunately coecov isn't showing the coverage on this)

parser.go
```diff
@@ -68,7 +68,7 @@ type Parser struct {
 	rawPlayers           map[int]*playerInfo                             // Maps entity IDs to 'raw' player info
 	additionalPlayerInfo [maxPlayers]common.AdditionalPlayerInformation  // Maps entity IDs to additional player info (scoreboard info)
 	modelPreCache        []string                                        // Used to find out whether a weapon is a p250 or cz for example (same id)
-	weapons              [maxEntities]common.Equipment                   // Used to remember what a weapon is (p250 / cz etc.)
+	weapons              map[int]*common.Equipment                       // Maps entity ID's to weapons. Used to remember what a weapon is (p250 / cz etc.)
```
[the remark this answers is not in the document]
markus-wa: Also I'm having a feeling we might want to (finally) move this to `gameState`. makes more sense in there I think.

what do you think?

sendtables/fake/property.go
```diff
@@ -13,7 +13,7 @@ type Property struct {
 	mock.Mock
 }
 
-// Namp is a mock-implementation of IProperty.Name().
+// Name is a mock-implementation of IProperty.Name().
```
markus-wa: Nice one! :smile:

parser.go
markus-wa: awesome work man, looks good!

go.mod
```diff
@@ -2,18 +2,29 @@ module github.com/markus-wa/demoinfocs-golang
 
 require (
 	github.com/dustin/go-heatmap v0.0.0-20180603032536-b89dbd73785a
+	github.com/go-gl/gl v0.0.0-20190320180904-bf2b1f2f34d7 // indirect
```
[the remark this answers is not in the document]
BestAwperEver: Not really sure why these lines ended up here in the first place, fixed

parser.go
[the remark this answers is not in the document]
BestAwperEver: Well considering many other `map[int]*something` fields inside `GameState` I would rather agree with you.

Worth noting that `Infernos()`, `GrenadeProjectiles()` and `Entities()` method do all return a reference to the underlying map without performing any copies. Maybe that should also be the way to treat `Weapons()`? Though I agree with you that this won't necessarily improve the overall performance I'm still voting for minimizing the amout of map-copying during the parsing process 😃

parser.go
```diff
@@ -157,6 +157,25 @@ func (p *Parser) Progress() float32 {
...
+	res := make(map[int]*common.Equipment)
+	for k, v := range p.weapons {
+		res[k] = v
+	}
+
+	return res
+}
+
+// WeaponTypeByEntityID returns the weapon type by its entity-ID.
+func (p *Parser) WeaponTypeByEntityID(entityID int) common.EquipmentElement {
```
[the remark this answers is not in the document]
BestAwperEver: As I mentioned a bit early, if we will go with the desing of all other entity getters inside `GameState`, this would be indeed redundant.

Anyhow, I just didn't think much of how it would look in terms of going in line with other API, I just added this after realizing that this is actually all the needed functionality and that I should have probably gone with this proposition initially instead of making all the weapons visible to the user but looks like it's a bit late for going back 😃 Now I see how it really doesn't fit there so yeah probably my little obsession with no-copy behaviour shouldn't be praised and it is better to make a more user-friendly interface.

datatables.go
BestAwperEver: Actually the only case when I faced this were in test when `OnCreate` were unintentional called on fake player entity because (if I'm not mistaken) its mocked method returned a non-mocked instance of `*st.Property` which have lead to this.

I believe that in reality this can only happen in some corrupted demos when a player entity would have a handle for unexisting entity so in this case we probably should create some error message (though I'm not very familiar with your way of creating an error during the parsing process so if I am the one to implement this behaviour I'd probably need some help). I'm not 100% sure but I've parsed about 10k demos using your parser and never had issues with.

parser.go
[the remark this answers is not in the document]
markus-wa: Alright, let's not do any copying then, but move it to `gameState` :smile:

parser.go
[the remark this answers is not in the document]
markus-wa: I think it's good that we expose the whole data, could be useful to someone. see above comment about leaving the copy-stuff out.

datatables.go
markus-wa: Fair point. we can raise a `ParserWarn` event if it ocurrs, like this: https://github.com/markus-wa/demoinfocs-golang/blob/7f5a03d810e41d0d78e8f7c2642e1d69c25957ce/game_events.go#L25

I prefer not aborting if it's not strictly needed, in the past the demo format changed occasionally where previous edge-cases suddenly became the norm, that way we can potentially avoid the need for a hotfix (though I'm pretty sure in this case this won't happen)

parser.go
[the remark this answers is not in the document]
BestAwperEver: 🥳

datatables.go
BestAwperEver: Right now I'm running the tests and it seems like in the very beggining of some demos there actually are these situations when a player has some weapon which hasn't beed binded prior to binding a player, so I'm having a mismatch with default.expected. What do you think we should do? May be leaving this as-is is actually not a bad idea because we don't really have any information about the weapon aside from its entity ID.

parser.go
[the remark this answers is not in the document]
BestAwperEver: Removed the method

BestAwperEver:
Moved `weapons` to `GameState`, removed the redundant `WeaponTypeByEntityID` method and added a parser warning if a player has non-existing weapon. Now with that warning regression tests fail, so we need to make a decision about how to treat this cases. If we leave it as-is, we should replace the expected parser output in test cases to include these new warnings.

markus-wa:
I'm interested to see if @reviewdog is working, now that I am unfortunately [forced to replace our dear friend](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e) @golangcibot

datatables.go
BestAwperEver: Just to clearify, does it mean that we should remove the `ParserWarn` or we should update `default.golden` instead? I suppose that it is the former as I don't see any lines with `ParserWarn` inside `default.expected`.

markus-wa:
hmm, travis isn't getting notified about this PR anymore, strange...

BestAwperEver:
> I am unfortunately [forced to replace our dear friend](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e) @golangcibot
Oh that's a pity. I just began to feel that we finally understand each other.

datatables.go
markus-wa: oh, right. Sorry I wrote it to read like we should remove `ParserWarn` but actually now that you say that it makes sense to update `default.golden`.

I thought the diff was the `UniqueID`s, but it's the `ParserWarn`s that are new, which is intended.

so you can run `go test -update` (I think)

BestAwperEver: Alright let's do it. Right now this is the only thing that doesn't allow tests to pass (everything is showing as OK if a remove the lines with `ParserWarn` dispatching) so I do hope that I will not break anything with it.

markus-wa:
something funny is going on with travis ...

can you squash and force-push something from your machine? or just something to re-trigger travis?

it's a bit strange but the commits here https://travis-ci.org/github/markus-wa/demoinfocs-golang/pull_requests don't line up with the commits inside this PR

BestAwperEver:
I fixed a typo in a method name, shoud it be enough to trigger travis?

markus-wa:
> should it be enough to trigger travis?

it *should* have been, but it didn't work :(

BestAwperEver:
I might misunderstand something, but it looks like Travis did the job but hasn't reported in this thread for some reason. Am I getting the problem correctly?
image

markus-wa:
> it looks like Travis did the job but hasn't reported in this thread for some reason

yes and no, the build has been done but the commit SHA is wrong. i'm not sure if travis not reporting back is a result of the wrong SHA or if that's another issue.

markus-wa:
btw, I saw there's now 2000 `ParserWarn`s in `default.golden`, let's remove the `ParserWarn` message and revert `default.golden`. No use in having so many warnings, it seems to be a completely normal case.

Sorry for the confusion, the self isolation must be getting to me :sweat_smile:

BestAwperEver:
Travis is now showing that he is working on the last commit with correct (matching to the last in this PR) SHA
https://travis-ci.org/github/markus-wa/demoinfocs-golang
image
Again not sure if i'm getting this correctly but may be this is just a matter of time 😄

markus-wa:
for me it's still the wrong SHA 
image

markus-wa:
Ah, that travis-pls branch was made by me by checking out your branch in my repo. but it's not doing what i had hoped

markus-wa:
ok, I have now migrated everything from the legacy travis.org platform to travis.com.

not sure if this will help but I'll give it some time to recover and see what's gonna happen

markus-wa:
I think we need to close and re-create this whole PR to fix the issues we're having with the CI services.

I think I've managed to fix things so it works for newly created PRs (did a quick test here: https://github.com/markus-wa/demoinfocs-golang/pull/194)

Noise should also be reduced with any new PRs as I've deleted CodeClimate (which was just producing duplicates and false positives) as well as DeepCode (which doesn't really support Go anyway)

BestAwperEver:
OK, going to create a new fresh PR then.

PR #197, opened 2020-03-27
changing the existing weapon instead of creating a new one

BestAwperEver:
Hotfix for the recent internal changes.

Without it we have many players with unknown weapons. Before the internal changes we had an actual array with fixed memory layout so every entity ID had a fixed pointer to an equipment, which doesn't hold now, so when we first have a player binded (with unknown weapons at the moment) and then we bind their weapons, the pointers we have in player's `rawWeapons` still point to the unknown weapons.

### DIFF
--- a/datatables.go
+++ b/datatables.go
@@ -483,7 +483,7 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement) 
 	if wep, ok := p.gameState.weapons[entityID]; ok {
 		currentOwner = wep.Owner
 
-		unassert.True(p.gameState.weapons[entityID].UniqueID() != 	0)
+		unassert.True(p.gameState.weapons[entityID].UniqueID() != 0)
 		// If we are here, we already have a player that holds this weapon
 		// so the zero-valued Equipment instance was already created in bindPlayer()
 		// In this case we should create a new Equipment instance with non-zero unique id
--- a/datatables.go
+++ b/datatables.go
@@ -483,18 +483,12 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement) 
 	if wep, ok := p.gameState.weapons[entityID]; ok {
 		currentOwner = wep.Owner
 
-		if p.gameState.weapons[entityID].UniqueID() == 0 {
-			// If we are here, we already have a player that holds this weapon
-			// so the zero-valued Equipment instance was already created in bindPlayer()
-			// In this case we should create a new Equipment instance with non-zero unique id
-			// but having the same memory address so player's rawWeapons would still have a pointer to it
-			*wep = *(common.NewEquipment(wepType))
-		} else {
-			// Something horrible has happened
-			p.eventDispatcher.Dispatch(events.ParserWarn{
-				Message: "trying to bind a new weapon but its entity is already bound",
-			})
-		}
+		unassert.True(p.gameState.weapons[entityID].UniqueID() != 	0)
+		// If we are here, we already have a player that holds this weapon
+		// so the zero-valued Equipment instance was already created in bindPlayer()
+		// In this case we should create a new Equipment instance with non-zero unique id
+		// but having the same memory address so player's rawWeapons would still have a pointer to it
+		*wep = *(common.NewEquipment(wepType))
 	} else {
 		p.gameState.weapons[entityID] = common.NewEquipment(wepType)
 	}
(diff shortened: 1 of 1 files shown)


datatables.go
```diff
@@ -484,7 +484,12 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement)
 		currentOwner = wep.Owner
 	}
 
-	p.gameState.weapons[entityID] = common.NewEquipment(wepType)
+	if currentOwner == nil {
+		p.gameState.weapons[entityID] = common.NewEquipment(wepType)
+	} else {
+		p.gameState.weapons[entityID].Weapon = wepType
+	}
+
```
markus-wa: this would lead to non-unique `Equipment.UniqueID()`s

We can try this instead:

```go
if wep, ok := p.gameState.weapons[entityID]; ok {
	currentOwner = wep.Owner

	// force new Equipment instance for UniqueID and entity bindings
	*wep = *(common.NewEquipment(wepType))
} else {
	p.gameState.weapons[entityID] = common.NewEquipment(wepType)
}
```

not exactly sexy but hey

BestAwperEver: Yeah I forgot that we don't set a unique ID for the weapon in this case so it will be 0. Probably we should also add a parser warning here if weapon's `UniqueID()` returns non-zero value. Will test it and commit if regression tests pass.

datatables.go
```diff
@@ -482,9 +482,23 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement)
...
+			// so the zero-valued Equipment instance was already created in bindPlayer()
+			// In this case we should create a new Equipment instance with non-zero unique id
+			// but having the same memory address so player's rawWeapons would still have a pointer to it
+			*wep = *(common.NewEquipment(wepType))
+		} else {
+			// Something horrible has happened
+			p.eventDispatcher.Dispatch(events.ParserWarn{
+				Message: "trying to bind a new weapon but its entity is already bound",
+			})
+		}
```
markus-wa: I've wanted to start using `unassert` a bit more over doing warnings that never happen.
This leads to better code coverage and we actually notice if an assertion fails, since `ParserWarn` is ignored.

assertions are disabled in release builds so this shouldn't interfere with anyone else's code.

I think the new way to go should be:
- if the regression set never goes down some path use `unassert` instead
- otherwise `ParserWarn` *might* be suitable

```suggestion
		unassert.True(p.gameState.weapons[entityID].UniqueID() != 	0)
		// If we are here, we already have a player that holds this weapon
		// so the zero-valued Equipment instance was already created in bindPlayer()
		// In this case we should create a new Equipment instance with non-zero unique id
		// but having the same memory address so player's rawWeapons would still have a pointer to it
		*wep = *(common.NewEquipment(wepType))
```

NOTE: I should probably add `unassert.NotSame()` which would be cleaner

markus-wa: PS: we should have probably used `unassert` in the last PR as well, but I forgot about it

datatables.go
```diff
@@ -482,9 +482,17 @@ func (p *Parser) bindWeapon(entity *st.Entity, wepType common.EquipmentElement)
 	var currentOwner *common.Player
 	if wep, ok := p.gameState.weapons[entityID]; ok {
 		currentOwner = wep.Owner
+
+		unassert.True(p.gameState.weapons[entityID].UniqueID() != 	0)
```
markus-wa: oops my bad.

```suggestion
		unassert.True(p.gameState.weapons[entityID].UniqueID() != 0)
```

BestAwperEver:
Why does Travis fail, exactly? Locally I ran regression tests on both `ParserWarn` and `unassert` variants and it seems to be OK, and also Travis passed here with `ParserWarn` as well

markus-wa:
travis failed on an unassert condition. to test these you have to run the tests with `-tags unassert_panic` (or run test scripts from `./bin/` which should do it automatically).

I did some debugging and took the liberty of pushing a fix to your branch: https://github.com/markus-wa/demoinfocs-golang/pull/197/commits/7c21a15c7e4910450b9f0a01e204fff5b2d82657

fingers crossed this works.

markus-wa:
I removed the `Something is clearly wrong here` as it seems to be a common case.

I accidentially also removed the other one, oops. will re-add it in a sec

markus-wa:
Merged - many thanks as always @BestAwperEver :raised_hands:

BestAwperEver:
No problem, was my mistake in the first place 😃

markus-wa:
> was my mistake in the first place

Only because I forced you to replace my ugly array with a map :sweat_smile:

PR #205, opened 2020-04-18
Bomb Damage Detection (+Fall Damage)

markus-wa:
image

Hey there @xavier-rodet & @BestAwperEver 
I think have finally found a way to detect bomb damage reliably.

Please note that this branch uses the v2 API, so you may need to upgrade.

I will merge this in a bit and will post an update on #169 when a new version with the fix is out.

### DIFF
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -51,8 +51,10 @@ func (p *parser) handleGameEvent(ge *msg.CSVCMsg_GameEvent) {
 }
 
 type gameEventHandler struct {
-	parser                 *parser
-	gameEventNameToHandler map[string]gameEventHandlerFunc
+	parser                  *parser
+	gameEventNameToHandler  map[string]gameEventHandlerFunc
+	userIDToFallDamageFrame map[int32]int
+	frameToRoundEndReason   map[int]events.RoundEndReason
 }
 
 func (geh gameEventHandler) dispatch(event interface{}) {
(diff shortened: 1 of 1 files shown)


BestAwperEver:
Hi @markus-wa ! Sorry for the late response, didn't have much time recently 😞 

The thing is, I'm not sure I will be able to review these changes anytime soon because of the `v2`-related stuff and the fact that I use reflection to make documents from the events provided by your library and store parsed data in the database. Yeah maybe I made a wrong decision back then and should have process each event manually with my own field structures and names but it's too late now 😃 I'm almost sure it would not be painful just for me to switch the major version but there are several more people using the output of the parser, and things like `Weapon.Weapon` and `Player.AdditionalPlayerInfo.*` sometimes go straight to the front-end without any changes at all.

With that being said, I'm hoping that I will eventually convince my team to run through the code and find all places which need to be changed but right now there are many critical issues with our project that need to be fixed before I could even imagine that we will have some time to switch to the `v2` branch.

By the time you read this you are probably somewhat interested in what "my team" is 😃 I did not mention it early (mainly because there is really not much to show off right now) but it's [scope.gg](https://scope.gg), and while we still have a looong way to be out of beta, there is already an (almost 😄) working 2D-replay which is the main reason I cannot switch to the `v2` branch right away.

markus-wa:
No worries, I figured a major version change might be a lot to ask, hence I just merged it after testing on my own 😄.



Also scope.gg looks really interesting, I had no idea! Would be cool to see it in action sometime 😄.

Mind if I add that to the 'projects using this lib' section?

BestAwperEver:
Yeah I think it would be suitable since at least in the replay section of our site others could already see the library in action

markus-wa:
> Yeah I think it would be suitable since at least in the replay section of our site others could already see the library in action.

where is that replay section? I can't see it somehow :sweat_smile: 

> analytical and advisory service for advanced CS:GO players

sounds good to me. I'll pop that in as description :smile:

BestAwperEver:
You could see it at "Replays" section of any uploaded match, [here](https://app.scope.gg/matches/3409697615577613027/replays) is an example. Although in the current state it runs at your display frequency so if you have a 144hz monitor it will be crazy fast 😃  

You don't actually have to register to watch the replay, ad-blocking the "sign in through steam" modal window is enough 😄

markus-wa:
I checked it out earlier this week and must say that looks really cool! 😄

BestAwperEver:
Thanks, glad to see that you like it!

PR #208, opened 2020-05-09
protobuf: add CS_UM_EndOfMatchAllPlayersData message

juliusmh:
The protobuf is from https://github.com/SteamDatabase/Protobufs/blob/e2103eff451173cd8b88b871ed8313cb80a15084/csgo/cstrike15_usermessages.proto.

Maybe you should consider pulling those protobufs directly into your project, maybe using submodule or similar.


markus-wa:
thanks @juliusmh , merged it :smile:

do you know what's exactly in `CCSUsrMsg_EndOfMatchAllPlayersData` by chance?

> Maybe you should consider pulling those protobufs directly into your project, maybe using submodule or similar.

I have considered this in the past but there is also a bunch of stuff in there that we aren't interested in, and we will manually have to update the submodule and re-generate the Go code anyway, so I decided against it.

Issue #244, opened 2020-11-05
PlayerHurt.HealthDamage could be improved


Julien2313:
**Higher goal**
`PlayerHurt.HealthDamage` returns the theoretical damages given, not the reals ones.

**Describe the solution you'd like**
Having a new fields with the real value. In example, if a AWP kill someone with HS with 15HP left, I'd like `PlayerHurt.HealthDamage` to returns 15, not 450

**Describe alternatives you've considered**
At the moment I get the real value by getting all the participants, find the victim, and if `PlayerHurt.Health == 0` I change my value of damage by the current health of the player (hopefully, this is never updated before the event)

Julien2313:
If you accept this update, I'd be glade to do it

markus-wa:
I think this would be reasonable enough to add if named and documented accordingly yeah. Bht I would like to better understand why you need this - mostly out of curiosity.

If you'd like to make a PR that'd be cool

Julien2313:
What bht means ? xD

I'm giving "impacts points" for players during a game. One of the way to get points is to deal damage. That's why I need to "real" value :D

markus-wa:
Bht is just a typo. Was supposed to be "but" 😄. And yeah that usecase makes sense, thanks for explaining

markus-wa:
closing as #247 implemented this, thanks for the contribution!

PR #247, opened 2020-11-07
event: add the real value of damage taken

Julien2313:
(#244)
I haven't found any test example for this case (I found in parser_test.go something I could work on, but it wouldn't have tested my changes)


pkg/demoinfocs/game_events.go
```diff
@@ -340,20 +340,39 @@ func (geh gameEventHandler) playerDeath(data map[string]*msg.CSVCMsg_GameEventKe
...
 func (geh gameEventHandler) playerHurt(data map[string]*msg.CSVCMsg_GameEventKeyT) {
+	userID := data["userid"].GetValShort()
+	player := geh.playerByUserID32(userID)
 	attacker := geh.playerByUserID32(data["attacker"].GetValShort())
+
 	wepType := common.MapEquipment(data["weapon"].GetValString())
-	userID := data["userid"].GetValShort()
 	wepType = geh.attackerWeaponType(wepType, userID)
 
+	health := int(data["health"].GetValByte())
```
markus-wa, line 350: this should be `GetValShort`

pkg/demoinfocs/events/events.go
```diff
@@ -367,7 +367,11 @@ type PlayerHurt struct {
 	WeaponString string            // Wrong for CZ, M4A1-S etc.
 	HealthDamage int
 	ArmorDamage  int
-	HitGroup     HitGroup
+	// damage taken are used to know the real value the player lost
+	// in example, if the player has been shot by an AWP with a HS and had 50 health left, this value would be 50, not ~420
+	HealthDamageTaken *int
+	ArmorDamageTaken  *int
```
markus-wa, line 373: these shouldn't have to be pointers

pkg/demoinfocs/game_events.go
```diff
@@ -340,20 +340,39 @@ func (geh gameEventHandler) playerDeath(data map[string]*msg.CSVCMsg_GameEventKe
...
-		HealthDamage: int(data["dmg_health"].GetValShort()),
-		ArmorDamage:  int(data["dmg_armor"].GetValByte()),
-		HitGroup:     events.HitGroup(data["hitgroup"].GetValByte()),
-		Weapon:       geh.getEquipmentInstance(attacker, wepType),
+		Player:            player,
+		Attacker:          attacker,
+		Health:            health,
+		Armor:             armor,
+		HealthDamage:      int(data["dmg_health"].GetValShort()),
+		ArmorDamage:       int(data["dmg_armor"].GetValByte()),
```
markus-wa, line 371: let's reuse the vars from above here

```suggestion
		HealthDamage:      healthDamage,
		ArmorDamage:       armorDamage,
```

pkg/demoinfocs/game_events.go
markus-wa: sorry, I misunderstood what's going on here

pkg/demoinfocs/game_events.go
markus-wa: ignore me

pkg/demoinfocs/events/events.go
Julien2313: true

markus-wa:
Thanks for the PR @Julien2313 ! - looks good overall, think we should just clean up some docs and remove the pointers :slightly_smiling_face:

Julien2313:
> 
> 
> Thanks for the PR @Julien2313 ! - looks good overall, think we should just clean up some docs and remove the pointers 🙂

I wrote too much doc ?! xD

markus-wa:
@Julien2313 not too much doc no, just about how it will be displayed in the generated doc. see https://github.com/markus-wa/demoinfocs-golang/pull/247#discussion_r519192771 (I did a ninja edit on that comment, so you might not have seen the docs bit about it)

markus-wa:
actually, I think it's not even about godoc, it's about IDE auto-complete. pkg.go.dev might have handled it ok

Julien2313:
> @Julien2313 not too much doc no, just about how it will be displayed in the generated doc. see [#247 (comment)](https://github.com/markus-wa/demoinfocs-golang/pull/247#discussion_r519192771) (I did a ninja edit on that comment, so you might not have seen the docs bit in there)

Hooo ok ok, got it

Julien2313:
> excluding over-damage 
So well said ! ha ha

pkg/demoinfocs/events/events.go
```diff
@@ -359,15 +359,17 @@ const (
...
+	Player            *common.Player // May be nil if the demo is partially corrupt (player is 'unconnected', see #156 and #172).
+	Attacker          *common.Player // May be nil if the player is taking world damage (e.g. fall damage) or if the demo is partially corrupt (player is 'unconnected', see #156 and #172).
+	Health            int
+	Armor             int
+	Weapon            *common.Equipment // May be EqUnknown for world-damage (falling / bomb).
+	WeaponString      string            // Wrong for CZ, M4A1-S etc.
+	HealthDamage      int
+	ArmorDamage       int
+	HealthDamageTaken int // HealthDamage excluding over-damage (e.g. if player has 5 health and is hit for 15 damage this would be 5 instead of 15)
+	ArmorDamageTaken  int // ArmorDamage excluding over-damage (e.g. if player has 5 armor and is hit for 15 armor damage this would be 5 instead of 15)
```
markus-wa, line 371: nice one for catching the typo in my suggestion :clap: :smile:

pkg/demoinfocs/game_events.go
```diff
@@ -340,20 +340,39 @@ func (geh gameEventHandler) playerDeath(data map[string]*msg.CSVCMsg_GameEventKe
...
+	player := geh.playerByUserID32(userID)
 	attacker := geh.playerByUserID32(data["attacker"].GetValShort())
+
 	wepType := common.MapEquipment(data["weapon"].GetValString())
-	userID := data["userid"].GetValShort()
 	wepType = geh.attackerWeaponType(wepType, userID)
 
+	health := int(data["health"].GetValByte())
+	armor := int(data["armor"].GetValByte())
+	healthDamageTaken := health
```
markus-wa, line 352: hmm, I think this is wrong. shouldn't it be `int(data["dmg_health"].GetValShort())` by default and if `data["health"].GetValByte() == 0` then it should be `player.Health()`.

but it should never be `data["health"].GetValByte()` - right?

markus-wa (commented):
hmm. I think there's an issue with the default value of `HealthDamageTaken` and `ArmorDamageTaken`

pkg/demoinfocs/game_events.go
Julien2313: Yes, you're right ! My bad

Julien2313:
I added a check if the value is > 100, let me know if it's ok for you or not

Julien2313:
Do you have somewhere an helper package, so I can add a max function max int ? To directly make the check at the assignation

pkg/demoinfocs/game_events.go
```diff
@@ -340,20 +340,50 @@ func (geh gameEventHandler) playerDeath(data map[string]*msg.CSVCMsg_GameEventKe
...
+		}
+	} else {
+		if healthDamageTaken > 100 {
+			healthDamageTaken = 100
+		}
+
+		if armorDamageTaken > 100 {
+			armorDamageTaken = 100
+		}
+	}
```
markus-wa, line 374: we should check this even if `player != nil`

markus-wa: probably best to just move it in front of that if statement

markus-wa:
> Do you have somewhere a helper package, so I can add a max function max int ? To directly make the check at the assignation.
(it will slow down the parser a little little bit as most of the time, player won't be nil, or I could let it like that)

I think it's fine as it is. it's a shame Go doesn't provide min/max functions for all data types in the stdlib, but I'd rather not add a new package or even function just for this.

markus-wa (approved):
LGTM :slightly_smiling_face: 

appreciate your efforts with all this, thanks a lot :pray:

Julien2313:
> 
> 
> LGTM 🙂
> 
> appreciate your efforts with all this, thanks a lot 🙏

You are welcome, that's a wonderful library ;)

PR #249, opened 2020-11-07
common: try to improve accuracy of Player.IsDucking()

### LINKED ISSUE #241: IsDucking() seems incorrect
**Describe the bug**
After examining a demo, IsDucking() seems to not be true when a player is ducked.
Instead, it is true for several frames proceeding the actual duck.

**To Reproduce**

[duck_bug.zip](https://github.com/markus-wa/demoinfocs-golang/files/5213380/duck_bug.zip)

**Expected behavior**
IsDucking() should be true when the player is actually in the ducking state.

**Library version**
I think v1.11.0

**Additional context**
Some digging:
https://forums.alliedmods.net/archive/index.php/t-60764.html
https://github.com/alliedmodders/sourcemod/blob/dea3ef70a2ce94e1cb86e0283c8e3babf7ab98ef/core/smn_entities.cpp#L2558
https://github.com/alliedmodders/sourcemod/blob/81dc80fbd14b7eba3c20bc22a60629fc874015de/plugins/include/entity_prop_stocks.inc#L131

markus-wa:
fixes #241 (hopefully)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,8 @@ linters:
     - lll
     - typecheck
     - gomnd
+    - exhaustivestruct
+    - godot
 issues:
   exclude-rules:
     # Exclude some linters from running on tests files.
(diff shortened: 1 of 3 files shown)


markus-wa:
:wave:  @87andrewh - sorry this took so long. could you give this a try to see if `IsDucking()` works better on this branch?

PR #269, opened 2021-03-10
Give some love to hostage maps

akiver:
Hi!

This PR allows to access hostages information and adds events related to hostage maps.

You can find a hostage demo here:
https://www85.zippyshare.com/v/ebimojQq/file.html

Thank you

### DIFF
--- a/pkg/demoinfocs/common/hostage.go
+++ b/pkg/demoinfocs/common/hostage.go
@@ -1,0 +1,65 @@
+package common
+
+import (
+	"github.com/golang/geo/r3"
+
+	st "github.com/markus-wa/demoinfocs-golang/v2/pkg/demoinfocs/sendtables"
+)
+
+// HostageState is the type for the various HostageStateXYZ constants.
+type HostageState byte
+
+// HostageState constants give information about hostages state.
+// e.g. being untied, picked up, rescued etc.
+const (
+	HostageStateIdle            HostageState = 0
+	HostageStateBeingUntied     HostageState = 1
+	HostageStateGettingPickedUp HostageState = 2
+	HostageStateBeingCarried    HostageState = 3
+	HostageStateFollowingPlayer HostageState = 4
+	HostageStateGettingDropped  HostageState = 5
+	HostageStateRescued         HostageState = 6
+	HostageStateDead            HostageState = 7
+)
+
+// Hostage represents a hostage.
+type Hostage struct {
+	Entity           st.Entity
+	demoInfoProvider demoInfoProvider
+}
+
+// Position returns the current position of the hostage.
+func (hostage *Hostage) Position() r3.Vector {
+	if hostage.Entity == nil {
+		return r3.Vector{}
+	}
+
+	return hostage.Entity.Position()
+}
+
+// State returns the current hostage's state.
+// e.g. being untied, picked up, rescued etc.
+// See HostageState for all possible values.
+func (hostage *Hostage) State() HostageState {
+	return HostageState(getInt(hostage.Entity, "m_nHostageState"))
+}
+
+// Health returns the hostage's health points.
+// ! On Valve MM matches hostages are invulnerable, it will always return 100 unless "mp_hostages_takedamage" is set to 1
+func (hostage *Hostage) Health() int {
+	return getInt(hostage.Entity, "m_iHealth")
+}
+
+// Leader returns the possible player leading the hostage.
+// Returns nil if the hostage is not following a player.
+func (hostage *Hostage) Leader() *Player {
+	return hostage.demoInfoProvider.FindPlayerByHandle(getInt(hostage.Entity, "m_leader"))
+}
+
+// NewHostage creates a hostage.
+func NewHostage(demoInfoProvider demoInfoProvider, entity st.Entity) *Hostage {
+	return &Hostage{
+		demoInfoProvider: demoInfoProvider,
+		Entity:           entity,
+	}
+}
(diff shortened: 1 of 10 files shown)


akiver:
@markus-wa
```
package should be `common_test` instead of `common` (testpackage)
```

Should I disable this linter?

pkg/demoinfocs/common/hostage.go
```diff
@@ -0,0 +1,65 @@
...
+	demoInfoProvider demoInfoProvider
+}
+
+// Position returns the current position of the hostage.
+func (hostage *Hostage) Position() r3.Vector {
+	if hostage.Entity == nil {
+		return r3.Vector{}
+	}
+
+	return hostage.Entity.Position()
```
markus-wa, line 37: Is this working as expected?
I'm worried this might be causing problems: https://github.com/markus-wa/demoinfocs-golang/blob/master/pkg/demoinfocs/sendtables/entity.go#L207-L242

I'll have a more detailed look later to confirm

pkg/demoinfocs/datatables.go
```diff
@@ -595,3 +596,23 @@ func (p *parser) bindGameRules() {
...
+		entity.OnDestroy(func() {
+			delete(p.gameState.hostages, entityID)
+		})
+
+		var state common.HostageState
+		entity.Property("m_nHostageState").OnUpdate(func(val st.PropertyValue) {
+			oldState := state
+			state = common.HostageState(val.IntVal)
+			if oldState != state {
+				p.eventDispatcher.Dispatch(events.HostageStateChanged{State: state, Hostage: p.gameState.hostages[entityID]})
```
markus-wa, line 614: should we also add the `oldState` as `PreviousState` to the event? might be helpful

markus-wa:
Thanks for this PR @akiver - that's some amazing first contribution! Great job on following existing patterns and good test coverage!

I think the reviewdog comment is correct, we should change test files to be in packages called `_test`

Other than that I just had two questions I put in some comments but I think once those are answered we should be good to merge this

:clap:

pkg/demoinfocs/common/hostage.go
akiver: I did a quick test by standing on the head of a hostage 😄 and it looks fine:

```
akiver (1721.302978515625000000000000, 753.940368652343750000000000, -95.837493896484375000000000)
Hostage (1720.000000000000000000000000, 752.000000000000000000000000, -159.875000000000000000000000)
```

pkg/demoinfocs/datatables.go
akiver: Added

akiver:
> I think the reviewdog comment is correct, we should change test files to be in packages called `_test`

If we move the tests in a package, non-exposed fields such as `demoInfoProvider` in the `Player` struct will need to be exposed.
What do you think?

pkg/demoinfocs/common/hostage.go
markus-wa: awesome, thanks for the quick check!

markus-wa:
> If we move the tests in a package, non-exposed fields such as demoInfoProvider in the Player struct will need to be exposed.
What do you think?

ahh :hankey: - you're right, let's not do that for now. I might have to go and re-jig those tests at some point (and make them run in parallel!)

pkg/demoinfocs/game_state.go
```diff
@@ -101,6 +102,11 @@ func (gs gameState) Rules() GameRules {
 	return gs.rules
 }
 
+// Hostages returns a map from entity-IDs to all hostages.
+func (gs gameState) Hostages() map[int]*common.Hostage {
```
markus-wa, line 106: Would it make more sense to return a slice here?

If we really want to return a map we should make a defensive copy of it, so nobody accidentally tries to mess with it
see https://github.com/markus-wa/demoinfocs-golang/blob/master/pkg/demoinfocs/game_state.go#L270-L293

but as there's probably not too many hostages on a map and I'm not sure consumers need to access them by entity ID id recommend returning a slice.

but let me know if you disagree :slightly_smiling_face:

markus-wa:
found something else, but I think that's all (promise)

pkg/demoinfocs/game_state.go
akiver: I agree, I now return a slice.

akiver:
> > If we move the tests in a package, non-exposed fields such as demoInfoProvider in the Player struct will need to be exposed.
> > What do you think?
> 
> ahh 💩 - you're right, let's not do that for now. I might have to go and re-jig those tests at some point (and make them run in parallel!)

I disabled the linter in the meantime 🙂

markus-wa:
> https://www85.zippyshare.com/v/ebimojQq/file.html

that link is not working for me @akiver :disappointed: 

But I think I'm happy with the code as is so I'll go ahead and merge it.
Could you re-upload that demo somewhere though? I might want to include it in the regression set.

markus-wa (approved):
great work! :partying_face:

akiver:
> > https://www85.zippyshare.com/v/ebimojQq/file.html
> 
> that link is not working for me @akiver 😞
> 
> But I think I'm happy with the code as is so I'll go ahead and merge it.
> Could you re-upload that demo somewhere though? I might want to include it in the regression set.

Sure, here is another link https://mega.nz/file/2B9mDRJa#6kbHaoGdjXF5ZVXWS8tJBj0Q3W1NfFFmfwljVe_fIFY

Thanks

markus-wa:
thanks @akiver - I've added the demo to the regression set and released v2.6.0 
https://github.com/markus-wa/demoinfocs-golang/releases/tag/v2.6.0

Issue #303, opened 2021-06-18
Solution for heatmap shrinkage - examples improvement


srjnm:
**Describe the bug**
The heatmap generated by [heatmap.go](https://github.com/markus-wa/demoinfocs-golang/blob/master/examples/heatmap/heatmap.go) gets shrinked if the size of dots are increased and goes out of the map bounds. This might be happening due to the bounding rectangle being too pixel perfect, which causes [go-heatmap](https://github.com/dustin/go-heatmap) to scale down the points.

**Solution**
I expanded the bounding rectangle by 16 units.
I had to come up with a brute-force solution out of frustration, but it fixed my problem quite well. The value 16 units most likely wouldn't fix your case, but hopefully it'll help you figure out your own value or come up with a better solution.

Original Dot Size: 15
My Dot Size: 30

Code:
lines 63-68: [heatmap.go](https://github.com/markus-wa/demoinfocs-golang/blob/master/examples/heatmap/heatmap.go)
Current-
```
       // Find bounding rectangle for points to get around the normalization done by the heatmap library
       r2Bounds := r2.RectFromPoints(points...)
       bounds := image.Rectangle{
       		Min: image.Point{X: int(r2Bounds.X.Lo), Y: int(r2Bounds.Y.Lo)},
       		Max: image.Point{X: int(r2Bounds.X.Hi), Y: int(r2Bounds.Y.Hi)},
       }
```

Changes-
```
       // Find bounding rectangle for points to get around the normalization done by the heatmap library
       r2Bounds := r2.RectFromPoints(points...)
       bounds := image.Rectangle{
       		Min: image.Point{X: int(r2Bounds.X.Lo - 16), Y: int(r2Bounds.Y.Lo - 16)},
       		Max: image.Point{X: int(r2Bounds.X.Hi + 16), Y: int(r2Bounds.Y.Hi + 16)},
       }
```

**Results**
Before - lines look messed up
![before](https://user-images.githubusercontent.com/54297985/122601332-c12fc980-d08e-11eb-83b2-fac753776d24.jpeg)
After - lines were perfect (:
![after](https://user-images.githubusercontent.com/54297985/122601358-cbea5e80-d08e-11eb-89aa-df6aa78f9b5e.jpeg)

markus-wa:
Interesting, well spotted @srjnm - what about `dotSize/2` instead of 16, considering your dot-size is 30 and 15/16px is basically the same? or is that much worse?

[...]

PR #304, opened 2021-06-18
Added padding to fix shrinkage

srjnm:
Related to issue #303 

Replaced `image.ZP` with `image.Point{}` as well, since it was outdated.


markus-wa:
Nice one, cheers! 🙏

Issue #334, opened 2022-04-01
Remove /assets and /pkg/demoinfocs/metadata


markus-wa:
The value proposition of this is superseded by https://github.com/saiko-tech/csgo-centrifuge

PR #346, opened 2022-06-04
examples: use radar-overviews.csgo.saiko.tech for radar images and offsets

markus-wa:
see #334

### DIFF
--- a/examples/common.go
+++ b/examples/common.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"strconv"
 
 	"github.com/golang/geo/r2"
 	"github.com/markus-wa/demoinfocs-golang/v2/pkg/demoinfocs/metadata"
@@ -60,6 +59,12 @@ func RedirectStdout(f func()) {
 	os.Stdout = old
 }
 
+type mapMetadata struct {
+	PosX  float64 `json:"pos_x,string"`
+	PosY  float64 `json:"pos_y,string"`
+	Scale float64 `json:"scale,string"`
+}
+
 // GetMapMetadata fetches metadata for a specific map version from
 // `https://radar-overviews.csgo.saiko.tech/<map>/<crc>/info.json`.
 // Panics if any error occurs.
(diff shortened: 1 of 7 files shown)


examples/common.go
```diff
@@ -46,3 +54,68 @@ func RedirectStdout(f func()) {
...
+
+	y, err := strconv.ParseFloat(mapInfo["pos_y"].(string), 64)
+	if err != nil {
+		return metadata.Map{}, fmt.Errorf("failed to get origin for Y coordinate: %v", err)
+	}
+
+	scale, err := strconv.ParseFloat(mapInfo["scale"].(string), 64)
+	if err != nil {
+		return metadata.Map{}, fmt.Errorf("failed to get scale: %v", err)
+	}
```
micvbang, line 93: Looks to me like it would make sense to declare the type of this upfront and let the JSON parser handle heavy lifting?

E.g.
```go
type mapInfo struct {
    PosX float32 `json:"pos_x"`
    PosY float32 `json:"pos_y"`
    Scale float32 `json:"scale"`
}
```

markus-wa: yeah, but the root looks like

```json
{
 "de_dust2": {...}
}
```

e.g.
- VDF: https://radar-overviews.csgo.saiko.tech/cs_agency/2230463619/info.txt
- JSON: https://radar-overviews.csgo.saiko.tech/cs_agency/2230463619/info.json

we want to keep that data as true to the original as possible, so getting rid of that root element /unesting is not really an option imo

We can probably 
- A: do some reflection foo-bar 
- or B: encode and re-decode the `de_dust2` part

to make this work properly, but that's not too nice either :cry: 

do you know a nice way of doing this?

markus-wa: > EDIT: ha, alright, now I managed to find the format and realized why you're doing it like this. Sorry :)

no worries :smile: - certainly looks like a code smell at first glance .. wish go was better at handling arbitrary data sometimes :sweat:

markus-wa: actually, `map[string]mapInfo` should work as type

https://stackoverflow.com/questions/18412126/golang-parse-a-json-with-dynamic-key

thanks for making me look at this again :smile:

micvbang: Nice! Yeah, that's quite a bit better :)

PR #367, opened 2022-11-15
Adding Recoil Index Property of Weapon

David-Durst:
This commit adds recoil index. This is easy to interpret: increasing recoil index means more recoil. Decreasing recoil index means player isn't holding fire button. 

Should I also add recoil vectors? I considered adding functions for `player.Entity.PropertyValueMust("localdata.m_Local.m_aimPunchAngle").VectorVal` and
`player.Entity.PropertyValueMust("localdata.m_Local.m_viewPunchAngle").VectorVal`, but decided against that because interpreting those vectors requires handling angle wrap around and scaling (see https://davidbdurst.com/blog/csknow_recoil.html). Thoughts? I wasn't sure if this amount of interpretation belongs in the parser.

### DIFF
--- a/pkg/demoinfocs/common/equipment.go
+++ b/pkg/demoinfocs/common/equipment.go
@@ -384,6 +384,18 @@ func (e *Equipment) AmmoReserve() int {
 	return val.IntVal
 }
 
+// RecoilIndex returns the weapon's recoil index
+func (e *Equipment) RecoilIndex() float32 {
+	if e.Entity == nil {
+		return 0
+	}
+
+	// if the property doesn't exist we return 0 by default
+	val, _ := e.Entity.PropertyValue("m_flRecoilIndex")
+
+	return val.FloatVal
+}
+
 // NewEquipment creates a new Equipment and sets the UniqueID.
 //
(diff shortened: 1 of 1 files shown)


markus-wa:
thanks @David-Durst :partying_face: 

> Should I also add recoil vectors? I considered adding functions for player.Entity.PropertyValueMust("localdata.m_Local.m_aimPunchAngle").VectorVal and
player.Entity.PropertyValueMust("localdata.m_Local.m_viewPunchAngle").VectorVal, but decided against that because interpreting those vectors requires handling angle wrap around and scaling (see https://davidbdurst.com/blog/csknow_recoil.html). Thoughts? I wasn't sure if this amount of interpretation belongs in the parser.

Yeah that would be a great addition.

Generally any feature that is "reliable" is welcome in the parser - I usually draw the line at things that are either demo-source specific (Valve MM, Faceit, Esportal - like knife round detection).
But as this is a general feature useful for any demo it would be great to have!

markus-wa:
Merged this as is for now, but if you wanted to do the the recoil vectors that would be very welcome :heart:

PR #371, opened 2022-12-26
feat: add neck hitbox

akiver:
Hi @markus-wa!

A tiny PR to add the hit group value for the neck.
It has been noticed in this [issue](https://github.com/akiver/CSGO-Demos-Manager/issues/598) that it's possible to have a hit group value set to `8`, which seems to correspond to the neck.

Thank you, and happy new year!


markus-wa:
Cheers @akiver - merged!

I will create a new release some time next year, until then just use @master if needed😊

Happy holidays!

PR #374, opened 2023-01-22
feat: improve POV demos support

akiver:
Hi!

This PR improves POV demos support.
The main problem was entities being fully re-created instead of updated when they left and re-entered PVS.
It was noticeable with player entities since they enter/leave PVS frequently with client demos.

Changes:
- Re-use existing entities when they enter PVS if possible based on their serial number
- Add int64 props decoder (got some with Danger mode demos)
- Use the proper prop in `player.Position()` (`cslocaldata.m_vecOrigin` vs `csnonlocaldata.m_vecOrigin`)
- Add a `parser.IsPOV` function
- Make `player.isAlive()` more accurate by checking the value of the prop `m_lifeState`
- Add more known game events
- Add a doc about game events

Running the following sample now outputs:
```
Player entity created GOTV 0
Player entity created TaRiK 3
Player entity created SiBaby 3
Player entity created stew 3
Player entity created valens 3
Player entity created unknown 3
Player entity created Metin 3
Player entity created sharof0444 3
Player entity created 1foxey 3
Player entity created sherpss 3
Player entity created cook 3
Player entity created Dan 0
Player entity created Seth 0
Player entity created Xander 0
```

Before this PR it would spam with entity creation.

```go
package main

import (
	"fmt"
	"os"

	demoinfocs "github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs"
	events "github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs/events"
	st "github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs/sendtables"
)

func main() {
	f, err := os.Open("../../test/cs-demos/set/POV-dust2-faceit-game_6-30.dem")
	checkError(err)

	defer f.Close()

	p := demoinfocs.NewParser(f)
	defer p.Close()

	header, err := p.ParseHeader()
	checkError(err)
	fmt.Println("Map:", header.MapName)

	p.RegisterEventHandler(func(events.DataTablesParsed) {
		p.ServerClasses().FindByName("CCSPlayer").OnEntityCreated(func(ent st.Entity) {
			player := p.GameState().Participants().FindByHandle(ent.ID())
			fmt.Println("Player entity created", player.Name, player.Team)
		})
	})

	err = p.ParseToEnd()
	checkError(err)
}

func checkError(err error) {
	if err != nil {
		panic(err)
	}
}
```

Thank you

### DIFF
--- a/pkg/demoinfocs/demoinfocs_test.go
+++ b/pkg/demoinfocs/demoinfocs_test.go
@@ -607,7 +607,7 @@ func assertGolden(tb testing.TB, assertions *assert.Assertions, testCase string,
 }
 
 func removePointers(s []byte) []byte {
-	r := regexp.MustCompile(`\(0x[\da-f]{10}\)`)
+	r := regexp.MustCompile(`\(0x[\da-f]+\)`)
 
 	return r.ReplaceAll(s, []byte("(non-nil)"))
 }
--- a/pkg/demoinfocs/net_messages.go
+++ b/pkg/demoinfocs/net_messages.go
@@ -3,6 +3,7 @@ package demoinfocs
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 
 	"github.com/markus-wa/ice-cipher-go/pkg/ice"
 	"google.golang.org/protobuf/proto"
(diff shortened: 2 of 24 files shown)


pkg/demoinfocs/demoinfocs_test.go
```diff
@@ -607,7 +607,7 @@ func assertGolden(tb testing.TB, assertions *assert.Assertions, testCase string,
 }
 
 func removePointers(s []byte) []byte {
-	r := regexp.MustCompile(`\(0x[\da-f]{10}\)`)
```
akiver, line 610: It can be 11 on arm64 arch

pkg/demoinfocs/net_messages.go
```diff
@@ -19,34 +20,30 @@ func (p *parser) handlePacketEntities(pe *msg.CSVCMsg_PacketEntities) {
 
 	r := bit.NewSmallBitReader(bytes.NewReader(pe.EntityData))
 
-	currentEntity := -1
-	for i := 0; i < int(pe.GetUpdatedEntries()); i++ {
-		currentEntity += 1 + int(r.ReadUBitInt())
-
-		cmd := r.ReadBitsToByte(2)
```
markus-wa, line 26: by replacing this with ReadBit we change the behaviour so that in certain cases the second bit is not read. I remember this caused a bug before which is why ReadBitsToByte(2) was introduced at some point.

Will need to check that this isn't bringing that bug back

pkg/demoinfocs/parser.go
```diff
@@ -190,6 +195,11 @@ func (p *parser) Progress() float32 {
 	return float32(p.currentFrame) / float32(p.header.PlaybackFrames)
 }
 
+// IsPOV indicates if the demo being parsed is a POV demo.
+func (p *parser) IsPOV() bool {
```
markus-wa, line 199: this may lead to some confusion when users call

`NewParser(demoReader).IsPOV()` as it will return false at first and only true after parsing a few frames.

Maybe it's better to only expose this information in the form of an event or something :thinking:

markus-wa:
thanks, @akiver - great work!

We'll need to iron out and test some details but looks really good :smile: 

I'll take a better look over the coming week as well but left some small comments for now

pkg/demoinfocs/parser.go
akiver: 👍 yeah I hesitated to add this detail in the comment.
I will replace it with an event similar to `TickRateInfoAvailable`.

akiver: I replaced it with an event `RecordingPlayerDetected`

markus-wa: Can we do `POVRecordingPlayerDetected` - might be more obvious to newcomers

akiver: Done

pkg/demoinfocs/net_messages.go
markus-wa: looks good :+1:

markus-wa (approved):

pkg/demoinfocs/net_messages.go
akiver: Thank you! I also parsed about 500 demos without any issues.

markus-wa:
merged and released v3.3.0 with these improvements https://github.com/markus-wa/demoinfocs-golang/releases/tag/v3.3.0

thanks again @akiver :heart:

PR #383, opened 2023-03-23
v4: Support Counter-Strike 2

markus-wa:


pkg/demoinfocs/s2_commands.go
```diff
@@ -0,0 +1,345 @@
...
+	return 0
+}
+
+func (p *parser) handleDemoPacket(pack *msgs2.CDemoPacket) {
+	r := bitread.NewSmallBitReader(bytes.NewReader(pack.Data))
+
+	ms := make([]pendingMessage, 0)
+
+	for r.ActualPosition() < len(pack.Data) {
+		t := int32(r.ReadUBitInt())
```
thecatontheflat, line 293: If I understood correctly, this one starts reading the buffer in 6 bits increments. Is it supposed to advance the offset in non-byte increments?

LaihoE: The reading is not "byte aligned", you just read bits from the stream. Also ReadUBitInt reads a varying number of bits, not always 6 bits.

markus-wa:
Game events are parsing ok so far, next up are entities 
image

ysomad:
@markus-wa Hi! Do you have some kind of a list which includes whats will be new and whats will be removed in a source2 parser?

PR #387, opened 2023-06-12
Fix NewEntityWithProperty (sendtables/fake)

supermaciz:
Hello

This PR fixes `NewEntityWithProperty`. This function can't work properly because the methods `PropertyValue` and `PropertyValueMust` need the name of the property as a parameter.


markus-wa:
thanks for the fix!

PR #393, opened 2023-08-04
add S2 method to Entity

BestAwperEver:
Hi! I encountered some errors while trying to parse a Source 2 demo file in `TeamStats` methods (since some properties have changed their names/hierarchy in S2), and decided that adding an `S2()` method to the `Entity` interface would be an acceptable solution. But then today I saw a PR #392 from @akiver, which solves the same problem just by looking for both the S1 and S2 property names in an `Entity`, and thus, if my change is accepted, those should be rewritten for consistency, so now I'm not so sure about the legitimacy of my suggestion. Anyway, I'm filing this PR as a place to discuss this. My take is that the `S2()` method and its usage clearly communicates to a user/developer that the properties are different specifically because of the engine, like this:
```go
// Score returns the current score of the team (usually 0-16 without overtime).
func (ts *TeamState) Score() int {
	var propName string
	if ts.Entity.S2() {
		propName = "m_iScore"
	} else {
		propName = "m_scoreTotal"
	}
	return getInt(ts.Entity, propName)
}
```

### DIFF
--- a/pkg/demoinfocs/common/common.go
+++ b/pkg/demoinfocs/common/common.go
@@ -136,12 +136,21 @@ func (ts *TeamState) Team() Team {
 
 // ID returns the team ID, this stays the same even after switching sides.
 func (ts *TeamState) ID() int {
+	if ts.Entity.S2() {
+		return int(getUInt64(ts.Entity, "m_iTeamNum"))
+	}
 	return getInt(ts.Entity, "m_iTeamNum")
 }
 
 // Score returns the current score of the team (usually 0-16 without overtime).
 func (ts *TeamState) Score() int {
-	return getInt(ts.Entity, "m_scoreTotal")
+	var propName string
+	if ts.Entity.S2() {
+		propName = "m_iScore"
+	} else {
+		propName = "m_scoreTotal"
+	}
+	return getInt(ts.Entity, propName)
 }
 
 // ClanName returns the team name (e.g. Fnatic).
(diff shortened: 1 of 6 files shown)


markus-wa:
Thanks for the PR and your thoughts on this.

I think adding something like an `S2()` func on sendtables turns this into a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction).
Admittedly the current abstraction is already leaky, as we store this info on PropertyValue as well (I want to clean this up before merging to main).

I think a better solution would be to add this to `NewPlayer(..., s2 bool)` - this way we're passing the info in from the top instead of it leaking from the dependency.

BestAwperEver:
I see your point. However, in that case, we would need to add this `s2` parameter not only to `Player`, but also to `TeamStats`, `Equipment`, `GrenadeProjectile` etc., essentially everywhere we use the `Entity` interface for retrieving properties by their names. This doesn't sound ideal to me, since it looks like duplicating a lot of what could be done in one place. However, this can be solved by creating something that would hold that `s2` property, which would be set once we read the filestamp, and by having a pointer to this "something" in every struct which also has an `Entity`.

markus-wa:
Good idea @BestAwperEver  - I think `demoInfoProvider` is a good place to put this and I think adding it to all missing structs would be a good call :+1:

BestAwperEver:
Closing this PR as I will be opening another with the `IsSource2` method being added to `demoInfoProvider`.

PR #394, opened 2023-08-04
feat: update header info from CDemoFileHeader and detect POV demos

akiver:
- Update header fields from the `CDemoFileHeader` message (should be the first message received)
- Detect POV demos the same way as CSGO demos
- Dispatch `PlayerInfo` events on `userinfo` string table updates

Thank you


markus-wa (approved):

markus-wa:
Thanks again @akiver :heart:

PR #395, opened 2023-08-07
add IsSource2 method to demoInfoProvider

BestAwperEver:
- add `IsSource2` method to `demoInfoProvider`, which calls `isSource2` method of `parser`
- add `demoInfoProvider` to `TeamState` (we will probably need to add it to other places as well)
- use `IsSource2` in `Player` and `TeamStats` to control which property of which entity should be used (there is still much to do in that regard)


markus-wa (approved):

markus-wa:
looks great, thanks!

PR #396, opened 2023-08-07
feat: weapons support

akiver:
This PR adds support for weapons accessible through `player.Weapons()` or `player.ActiveWeapon()`.

I don't use the same technique as for Source 1 to retrieve weapons entities and their `EquipmentType` mapping to populate `gameState.weapons`.
Instead of using ServerClass names and/or 3D model names, I use the prop `m_iItemDefinitionIndex`.  
This prop corresponds to the unique equipment index. They are located in the game file `scripts/items/items_game.txt` and are never changed.
I think this way feels less hacky than checking for model/server class names. The only downside is that if a new weapon is added to the game, we must add it to our internal `map`.
Also I'm not sure we could use the same technique with Source 2 demos as the string table `modelprecache` is not present with Source 2 demos.

Notes:
- An entity is considered a weapon entity if the props `m_iItemDefinitionIndex` and `m_iClip1` exist (there may be a better way?)
- I added an `EquipmentType` for all missing equipment indexes from the latest CS:GO file, I don't have access to CS2 but it looks like indexes didn't change
- The prop `m_iPrimaryAmmoType` used to retrieve `AmmoReserve` is not present, but there is a new prop `m_pReserveAmmo` as a replacement
- I didn't find a replacement for `AmmoType` (the prop `LocalWeaponData.m_iPrimaryAmmoType` doesn't exist). I think it's ok since it was mainly used only to compute `AmmoReserve`
- The parser emits a warning if an equipment index is unknown
- I think we could use the same index-based approach with source 1 demos

Test code: 
```go
package main

import (
	"errors"
	"fmt"
	"os"
	"time"

	demoinfocs "github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs"
	"github.com/markus-wa/demoinfocs-golang/v3/pkg/demoinfocs/events"
)

func main() {
	var err error
	f, err := os.Open("./faze-vs-legends-m3-dust2.dem")
	checkError(err)

	defer f.Close()

	p := demoinfocs.NewParser(f)
	defer p.Close()

	_, err = p.ParseHeader()
	checkError(err)

	p.RegisterEventHandler(func(e events.Kill) {
		if e.Killer == nil {
			return
		}

		activeWeapon := e.Killer.ActiveWeapon()
		if activeWeapon != nil {
			fmt.Printf("killer:%s event weapon:%s active weapon:%s\n", e.Killer.Name, e.Weapon, activeWeapon)
		}

		for _, pl := range p.GameState().Participants().Playing() {
			weapons := pl.Weapons()
			if len(weapons) > 0 {
				fmt.Printf("----- %s (%d weapons) -----\n", pl.Name, len(weapons))
				for _, wp := range weapons {
					fmt.Printf("%s class:%d reserve:%d magazine:%d zoom:%d\n", wp.Type, wp.Class(), wp.AmmoReserve(), wp.AmmoInMagazine(), wp.ZoomLevel())
				}
				fmt.Println("---------------------")
			}
		}
	})

	now := time.Now()
	err = p.ParseToEnd()
	if errors.Is(err, demoinfocs.ErrUnexpectedEndOfDemo) {
		fmt.Println("Parsing done with unexpected EOF")
	} else {
		checkError(err)
	}
	fmt.Println(time.Since(now))
}

func checkError(err error) {
	if err != nil {
		panic(err)
	}
}

```

I also run a "stress" test where I log weapons every frame without issue:
```go
p.RegisterEventHandler(func(e events.FrameDone) {
		for _, pl := range p.GameState().Participants().Playing() {
			weapons := pl.Weapons()
			if len(weapons) > 0 {
				fmt.Printf("----- %s (%d weapons) -----\n", pl.Name, len(weapons))
				for _, wp := range weapons {
					fmt.Printf("%s class:%d reserve:%d magazine:%d zoom:%d\n", wp.Type, wp.Class(), wp.AmmoReserve(), wp.AmmoInMagazine(), wp.ZoomLevel())
				}
				fmt.Println("---------------------")
			}
		}
	})
```

Thank you

### DIFF
--- a/pkg/demoinfocs/common/equipment.go
+++ b/pkg/demoinfocs/common/equipment.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"math/rand"
+	"reflect"
 	"strings"
 
 	"github.com/oklog/ulid/v2"
@@ -97,13 +98,27 @@ const (
 
 	// Equipment
 
-	EqZeus      EquipmentType = 401
-	EqKevlar    EquipmentType = 402
-	EqHelmet    EquipmentType = 403
-	EqBomb      EquipmentType = 404
-	EqKnife     EquipmentType = 405
-	EqDefuseKit EquipmentType = 406
-	EqWorld     EquipmentType = 407
+	EqZeus                     EquipmentType = 401
+	EqKevlar                   EquipmentType = 402
+	EqHelmet                   EquipmentType = 403
+	EqBomb                     EquipmentType = 404
+	EqKnife                    EquipmentType = 405
+	EqDefuseKit                EquipmentType = 406
+	EqWorld                    EquipmentType = 407
+	EqZoneRepulsor             EquipmentType = 408
+	EqShield                   EquipmentType = 409
+	EqHeavyAssaultSuit         EquipmentType = 410
+	EqNightVision              EquipmentType = 411
+	EqHealthShot               EquipmentType = 412
+	EqTacticalAwarenessGrenade EquipmentType = 413
+	EqFists                    EquipmentType = 414
+	EqBreachCharge             EquipmentType = 415
+	EqTablet                   EquipmentType = 416
+	EqAxe                      EquipmentType = 417
+	EqHammer                   EquipmentType = 418
+	EqWrench                   EquipmentType = 419
+	EqSnowball                 EquipmentType = 420
+	EqBumpMine                 EquipmentType = 421
 
 	// Grenades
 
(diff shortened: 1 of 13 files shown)


pkg/demoinfocs/common/equipment.go
```diff
@@ -339,13 +357,30 @@ func (e *Equipment) AmmoInMagazine() int {
 		return -1
 	}
 
-	// need to subtract 1 as m_iClip1 is nrOfBullets + 1
-	return val.IntVal - 1
+	isSource1 := reflect.TypeOf(val.Any).Kind() == reflect.Int
```
akiver, line 360: It's a `uint32` with Source 2, this is a quick way to support both at the same time without a crash

pkg/demoinfocs/game_state.go
```diff
@@ -174,7 +175,15 @@ func (gs gameState) PlayerResourceEntity() st.Entity {
 	return gs.playerResourceEntity
 }
 
-func entityIDFromHandle(handle uint64) int {
+func entityIDFromHandle(handle uint64, isS2 bool) int {
+	if isS2 {
```
akiver, line 179: I needed to distinguish Source 1 and 2 here since constants are different.
The easier but not nice solution I found is to inject a boolean here and as a side effect pass the `isSource2` function from `parser` to the struct `participants`.

pkg/demoinfocs/common/equipment.go
```diff
@@ -339,13 +357,30 @@ func (e *Equipment) AmmoInMagazine() int {
...
+	if isSource1 {
+		// need to subtract 1 as m_iClip1 is nrOfBullets + 1
+		return val.Int() - 1
+	}
```
markus-wa, line 364:
```suggestion
	s1Ammo, isSource1 := val.Any.(int)
	if isSource1 {
		// need to subtract 1 as m_iClip1 is nrOfBullets + 1
		return s1Ammo - 1
	}
```

might be a bit nicer

markus-wa: either way, long term @BestAwperEver's #393 or followup thereof can be used in the future

markus-wa:
would definitely be nice to switch source 1 to item def index as well, but it would need to be backwards compatible (for S2 I don't mind if it is or isn't compatible)

markus-wa:
great work again on the whole PR, just had one small comment :smile:

pkg/demoinfocs/common/equipment.go
akiver: Changed!

markus-wa (approved):

PR #401, opened 2023-08-15
get player m_bPawnIsAlive property when pawn entity is nil

BestAwperEver:
At least sometimes `p.PlayerPawnEntity()` returns `nil`, so, in that case, we could still fall back to the `m_bPawnIsAlive` player property. I can't say anything about the accuracy but at least we don't crash :)


markus-wa (approved):

markus-wa:
thanks!

PR #405, opened 2023-08-24
fix: possible crash player props

akiver:
Small fixes to avoid possible crashes.

### DIFF
--- a/pkg/demoinfocs/common/player.go
+++ b/pkg/demoinfocs/common/player.go
@@ -156,9 +156,12 @@ This isn't very conclusive but it looks like IsFlashed isn't super reliable curr
 
 // Used internally to set the active weapon, see ActiveWeapon()
 func (p *Player) activeWeaponID() int {
-	pawnEntity := p.PlayerPawnEntity()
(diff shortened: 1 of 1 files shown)


pkg/demoinfocs/common/player.go
```diff
@@ -454,7 +458,12 @@ func (p *Player) PositionEyes() r3.Vector {
 	}
 
 	pos := p.Position()
-	pos.Z += float64(p.Entity.PropertyValueMust("localdata.m_vecViewOffset[2]").FloatVal)
+	if p.demoInfoProvider.IsSource2() {
+		// TODO Find out where we can find the offset in Source 2 demos
+		return pos
```
akiver, line 463: Like the velocity, I'm not sure it's available now. I didn't find a replacement.

markus-wa: can you open a issue for these as a bug?

akiver: done https://github.com/markus-wa/demoinfocs-golang/issues/406

PR #431, opened 2023-10-05
Exposing detailed knife details

Falderebet:
### Why this PR is needed
As of now it is not possible to see what "type" of knife players have equipped as they are all maped to EqKnife. This PR is an attempt to make that possible.

### What this PR adds
* New EquipmentTypes for the different knives.
* Refactored MapEquipment as it no longer needs to handle "edge" cases that knives where previously.


pkg/demoinfocs/common/equipment.go
```diff
@@ -276,13 +336,8 @@ const weaponPrefix = "weapon_"
...
 
-	var wep EquipmentType
-	if strings.Contains(eqName, "knife") || strings.Contains(eqName, "bayonet") {
-		wep = EqKnife
-	} else {
-		// If the eqName isn't known it will be EqUnknown as that is the default value for EquipmentType
-		wep = eqNameToWeapon[eqName]
-	}
+	// If the eqName isn't known it will be EqUnknown as that is the default value for EquipmentType
+	wep := eqNameToWeapon[eqName]
```
markus-wa, line 340: hmm, this doesn't really scale well for when they add new knifes - we would need to update the parser every time. and worse, our consumers would need to upgrade, otherwise they will get `EqUnknown` for the new knife, which would be really hard to debug.

I suggest that we instead expose the raw `eqName` somewhere upstream in the call chain

Falderebet: That is a good point, I will take a look at it!

Falderebet: I don't think there is a good way to handle valve adding new knives. However, I think there are some options:

1. Exposing the weapon "code"/itemindex - eg. we give the consumers the possibility to see what code the EqUnkown has.
    - Pros: 
        * This technically means that consumers can still handle new knives even before the parser is updated.
   - Cons:
       * It will still to a degree be hard for consumers to debug, as they only get a weapon code and nothing "human readable". 
       * Parser will still need to be updated to handle new knives with EqTypes and so forth. 

2. Exposing a more raw weapon string. 
    - Pros:
        * Makes it more human readable for consumers to debug.
    - Cons:
        * The parser needs to be updated before consumers can add new knives.
        
After writing this out I find myself gravitating more towards option 1. As the raw weapon string is not something coming from the engine but something we attach from the item index.

akiver: also why not keeping `strings.Contains(eqName, "knife")` as a fallback when `wep` is unknown?
i think it's better to have `weapon_knife` instead of "Unknown" when we know it's a knife

Falderebet: > also why not keeping `strings.Contains(eqName, "knife")` as a fallback when `wep` is unknown? i think it's better to have `weapon_knife` instead of "Unknown" when we know it's a knife

My thought was that the parser defines what a knife "is". Therefore it would be an error in the parser if the EqName did not map to an EqType.

I can add it back, I guess it is nice to have if new knives are added and the eqname/eqtype map is not updated.

markus-wa: Actually - I don't think we can add new `EqKnifeXYZ` types anyway as it's not backwards compatible and would lead to confusion (e.g. if the user wants to check if a kill was a knife kill, using `EqKnife` is more straight forward than checking all cases.

So we should expose this as additional data, not as replacement.

So we'll need another way of solving this - e.g. exposing the input string higher up

Falderebet: What about the idea of adding a field to the equipment struct with additional detailes about the weapon/equipment? 

Let me know if there is something I can do. :)

markus-wa (changes requested):
as discussed in comments

pkg/demoinfocs/common/equipment.go
```diff
@@ -118,6 +118,25 @@ const (
...
+	EqKnifeButterfly           EquipmentType = 431
+	EqKnifePush                EquipmentType = 432
+	EqKnifeCord                EquipmentType = 433
+	EqKnifeCanis               EquipmentType = 434
+	EqKnifeUrsus               EquipmentType = 435
+	EqKnifeGypsyJackknife      EquipmentType = 436
+	EqKnifeOutdoor             EquipmentType = 437
+	EqKnifeStiletto            EquipmentType = 438
+	EqKnifeWidowmaker          EquipmentType = 439
+	EqKnifeSkeleton            EquipmentType = 440
```
markus-wa, line 139: the `RawEquipment` part looks good now - but this part will need to be removed as it's not backwards compatible as mentioned in the other thread.

So all we should do for now is expose `RawEquipment`.

Maybe we can have a separate type `KnifeType` and constants `KnifeTypeXYZ` with a map `KnifeTypes map[string]KnifeType` which includes all known knife types to still provide some kind of mapping.

The good thing about this is that if a new knife comes out the users can just add it to the map themselves,. as the map will be mutable.

I think this would be the best middleground

pkg/demoinfocs/common/equipment.go
```diff
@@ -204,6 +223,28 @@ func initEqNameToWeapon() {
...
+	eqNameToWeapon["knife_butterfly"] = EqKnifeButterfly
+	eqNameToWeapon["knife_push"] = EqKnifePush
+	eqNameToWeapon["knife_cord"] = EqKnifeCord
+	eqNameToWeapon["knife_canis"] = EqKnifeCanis
+	eqNameToWeapon["knife_ursus"] = EqKnifeUrsus
+	eqNameToWeapon["knife_gypsy_jackknife"] = EqKnifeGypsyJackknife
+	eqNameToWeapon["knife_outdoor"] = EqKnifeOutdoor
+	eqNameToWeapon["knife_stiletto"] = EqKnifeStiletto
+	eqNameToWeapon["knife_widowmaker"] = EqKnifeWidowmaker
+	eqNameToWeapon["knife_skeleton"] = EqKnifeSkeleton
```
markus-wa, line 246: same here, would need to be moved to the new map I wrote about

pkg/demoinfocs/common/equipment.go
```diff
@@ -267,6 +308,25 @@ func initEqElementToName() {
...
+	eqElementToName[EqKnifeButterfly] = "Butterfly Knife"
+	eqElementToName[EqKnifePush] = "Push Knife"
+	eqElementToName[EqKnifeCord] = "Cord Knife"
+	eqElementToName[EqKnifeCanis] = "Canis Knife"
+	eqElementToName[EqKnifeUrsus] = "Ursus Knife"
+	eqElementToName[EqKnifeGypsyJackknife] = "Gypsy Jackknife"
+	eqElementToName[EqKnifeOutdoor] = "Outdoor Knife"
+	eqElementToName[EqKnifeStiletto] = "Stiletto Knife"
+	eqElementToName[EqKnifeWidowmaker] = "Widowmaker Knife"
+	eqElementToName[EqKnifeSkeleton] = "Skeleton Knife"
```
markus-wa, line 329: this would have to be in a new map `knifeTypeToName` - don't forget to implement the `String()` method on `KnifeType`

pkg/demoinfocs/common/equipment.go
```diff
@@ -543,23 +604,23 @@ var EquipmentIndexMapping = map[uint64]EquipmentType{
...
+	515: EqKnifeButterfly,           // weapon_knife_butterfly
+	516: EqKnifePush,                // weapon_knife_push
+	517: EqKnifeCord,                // weapon_knife_cord
+	518: EqKnifeCanis,               // weapon_knife_canis
+	519: EqKnifeUrsus,               // weapon_knife_ursus
+	520: EqKnifeGypsyJackknife,      // weapon_knife_gypsy_jackknife
+	521: EqKnifeOutdoor,             // weapon_knife_outdoor
+	522: EqKnifeStiletto,            // weapon_knife_stiletto
+	523: EqKnifeWidowmaker,          // weapon_knife_widowmaker
+	525: EqKnifeSkeleton,            // weapon_knife_skeleton
```
markus-wa, line 625: these would need to be removed as well

markus-wa, line 625: these would need to be put in a new map as well `KnifeTypeIndexMapping`

markus-wa:
thanks for the changes @Falderebet - sorry this is taking a while, but I think it's important to get this right to prevent issues in the future - thanks for your patience here!

markus-wa:
stale, superseded by #683

PR #440, opened 2023-10-13
Add otherdeath event handling

Falderebet:
### Why this is needed
As of now it is not possible to gather information about chickens killed.

### What this pr adds
Adds event handling of the "other_death" server event and dispatches an OtherEvent struct.

### DIFF
--- a/pkg/demoinfocs/events/events.go
+++ b/pkg/demoinfocs/events/events.go
@@ -504,6 +504,15 @@ type RankUpdate struct {
 	Player     *common.Player // may be nil if the player has already disconnected
 }
 
+// OtherDeath signals that there has occured a death of something that is not a player.
+// For example chickens.
+type OtherDeath struct {
+	Killer        *common.Player // May be nil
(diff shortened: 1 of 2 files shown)


markus-wa:
thanks! that's a great idea :smile: 

there's quite a bit more data in this event - would you be able to add it as well (through smoke, penetrated etc.)?

Falderebet:
> thanks! that's a great idea 😄
> 
> there's quite a bit more data in this event - would you be able to add it as well (through smoke, penetrated etc.)?

Yes, I can do that. :)

pkg/demoinfocs/events/events.go
Falderebet, line 1: I have left out the headshot bool - as far as i know none of the otherdeath entities can be headshotted, but i might be wrong.

akiver:
Thank you!

Ok for me, tested on a bunch of csgo/cs2 demos 🐔

markus-wa:
great work, thanks!

PR #444, opened 2023-10-15
inferno/Fires: add List() to return raw list of fires

micvbang:
This PR lets users to extract the raw list of individual fires, allowing them to perform custom calculations on the raw data. This can be very useful when e.g. wishing to visualize the data with higher fidelity than convex hull, and for doing statistics


akiver:
Thank you!

PR #452, opened 2023-10-21
grenade trajectories: collect timings of each location

micvbang:
This PR adds the possibility to track the timings of a grenade trajectory. This can be useful if you e.g. want to be able to show the trajectory as-it-happens during a replay.

I didn't want to make an API breaking change, so I added it in a separate list that tracks the contents of `GrenadeProjectile.Trajectory`. 

I added the `PointInTime` struct instead of just logging the frame id or time in seconds s.t. the user can get whichever of the two values they want. This struct could alternatively be called e.g. `PointInDemo`, `Timing`, or something else

### DIFF
--- a/pkg/demoinfocs/common/common.go
+++ b/pkg/demoinfocs/common/common.go
@@ -67,9 +67,13 @@ func (h *DemoHeader) FrameTime() time.Duration {
 type GrenadeProjectile struct {
 	Entity         st.Entity
 	WeaponInstance *Equipment
-	Thrower        *Player     // Always seems to be the same as Owner, even if the grenade was picked up
-	Owner          *Player     // Always seems to be the same as Thrower, even if the grenade was picked up
-	Trajectory     []r3.Vector // List of all known locations of the grenade up to the current point
+	Thrower        *Player // Always seems to be the same as Owner, even if the grenade was picked up
+	Owner          *Player // Always seems to be the same as Thrower, even if the grenade was picked up
+
+	// Deprecated: use Trajectory2 instead
+	Trajectory []r3.Vector // List of all known locations of the grenade up to the current point
+
+	Trajectory2 []TrajectoryEntry // List of all known locations and the point in time of the grenade up to the current point
 
 	// uniqueID is used to distinguish different grenades (which potentially have the same, reused entityID) from each other.
 	uniqueID int64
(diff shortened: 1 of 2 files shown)


pkg/demoinfocs/common/common.go
```diff
@@ -65,11 +65,12 @@ func (h *DemoHeader) FrameTime() time.Duration {
...
-	WeaponInstance *Equipment
-	Thrower        *Player     // Always seems to be the same as Owner, even if the grenade was picked up
-	Owner          *Player     // Always seems to be the same as Thrower, even if the grenade was picked up
-	Trajectory     []r3.Vector // List of all known locations of the grenade up to the current point
+	Entity           st.Entity
+	WeaponInstance   *Equipment
+	Thrower          *Player       // Always seems to be the same as Owner, even if the grenade was picked up
+	Owner            *Player       // Always seems to be the same as Thrower, even if the grenade was picked up
+	Trajectory       []r3.Vector   // List of all known locations of the grenade up to the current point
+	TrajectoryTiming []PointInTime // List of points in time for known locations of the grenade up to the current point
```
markus-wa, line 73: Maybe we should deprecate `Trajectory` and add a new field `Trajectory2` of type `TrajectoryEntry` and then we can add `r3.Vector` and Frame + Time to that struct?

micvbang: Yeah sure, that could work as well!

micvbang: Something like this?

pkg/demoinfocs/common/common.go
```diff
@@ -67,9 +67,13 @@ func (h *DemoHeader) FrameTime() time.Duration {
...
 	Entity         st.Entity
 	WeaponInstance *Equipment
-	Thrower        *Player     // Always seems to be the same as Owner, even if the grenade was picked up
-	Owner          *Player     // Always seems to be the same as Thrower, even if the grenade was picked up
-	Trajectory     []r3.Vector // List of all known locations of the grenade up to the current point
+	Thrower        *Player // Always seems to be the same as Owner, even if the grenade was picked up
+	Owner          *Player // Always seems to be the same as Thrower, even if the grenade was picked up
+
+	// Deprecated: use Trajectory2 instead
+	Trajectory []r3.Vector // List of all known locations of the grenade up to the current point
```
markus-wa, line 74:
```suggestion
	Trajectory []r3.Vector // Deprecated, use `Trajectory2` instead
```

markus-wa (approved):

pkg/demoinfocs/common/common.go
markus-wa: looks great!

markus-wa:
Thanks for the PR @micvbang  :heart:

PR #453, opened 2023-10-21
use m_firePositions property in compatible demos

BestAwperEver:
In the last 24 hours, I've been observing many demos that don't have `m_fire{X,Y,Z}Delta.XXXX` properties, and instead have `m_firePositions.XXXX` (older demos lack this property). This PR attempts to use the new property if the entity has it. I'm not 100% sure about the correctness of the data (see image: the CT died from the fire but the position of the death looks a bit off), but it looks good enough for now and is certainly better than crashing :)
image


micvbang:
Nice find! I can confirm that this is an issue and that your patch fixes it

akiver (approved):

PR #467, opened 2023-11-08
Using map for handlers instead of arrays

BestAwperEver:
In [this match](https://www.faceit.com/en/cs2/room/1-6908fbf4-d043-4ae4-9b40-16010def50b0) `xddxxxddxdx` disconnects during the match, and `BOT Rebel` in his place takes control over a pawn entity that previously belonged to `tiasaid`. Then `xddxxxddxdx` reconnects (as a side note, their nickname changes to `.`, which is their steam nickname -- I'm yet to investigate the case), which disconnects the bot, but because the pawn entity now has two `OnDestroy` handlers, `tiasaid` also gets marked as disconnected for the rest of the match. This also has other side-effects, because other properties also have two `OnUpdate` handlers.

This PR fixes the behaviour by replacing array of handlers with maps and adding an option to use a fixed handler id, which helps with overriding handlers rather than adding new.

I think this deserves having its own test suit, but probably due to inexperience I failed to wrap my head around calling handlers from the mocked entity on mocked property update :disappointed:

### DIFF
--- a/pkg/demoinfocs/datatables.go
+++ b/pkg/demoinfocs/datatables.go
@@ -526,6 +526,7 @@ func (p *parser) bindNewPlayerControllerS2(controllerEntity st.Entity) {
 
 func (p *parser) bindNewPlayerPawnS2(pawnEntity st.Entity) {
 	var prevControllerHandle uint64
+	handlerId := int64(pawnEntity.ID())
 
 	pawnEntity.Property("m_hController").OnUpdate(func(controllerHandleVal st.PropertyValue) {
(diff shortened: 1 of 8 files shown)


akiver:
Hi @BestAwperEver I took a look at the issue and opened an alternative https://github.com/markus-wa/demoinfocs-golang/pull/468
Could you please take a look at the PR and do some tests?

BestAwperEver:
Yep, that works as well!

markus-wa:
Thanks for the report @BestAwperEver - fixed in #468 instead to avoid changing the public API

PR #468, opened 2023-11-11
fix: retrieve and update correct player after disconnection/bot take over

akiver:
Alternative to fix https://github.com/markus-wa/demoinfocs-golang/pull/467.

This PR moves the logic inside the `OnUpdate` callback of `m_hController` outside to avoid calling pawn entity properties lifecycle when we should not.
The pawn entity props that we need to listen now retrieve the player associated with the pawn entity from the current `m_hController` prop value.


markus-wa:
nice work! I agree this seems like a good fix for #467

markus-wa (approved):

PR #475, opened 2023-11-18
fix: users not found in game events with new demos

akiver:
Since the last update, game events contain "user slot" rather than "user id" that come from the userinfo table.
Tested on a bunch of valve and faceit demos.

fix https://github.com/markus-wa/demoinfocs-golang/issues/476


pkg/demoinfocs/game_events.go
```diff
@@ -119,7 +119,17 @@ func (geh gameEventHandler) gameState() *gameState {
 }
 
 func (geh gameEventHandler) playerByUserID(userID int) *common.Player {
-	return geh.gameState().playersByUserID[userID]
+	player := geh.gameState().playersByUserID[userID]
+	if player != nil || !geh.parser.isSource2() {
```
akiver, line 123: I don't change it for Source 1 demos because in case of damages/kills done by the "world" it would return the GOTV entity (id `0`) instead of nil.

In Source 2 the CSTV entity ID is not always 0 and may have `SourceTV` or `DemoAutoRecorder` as a name.

markus-wa (approved):

Issue #458, opened 2023-10-25
Apparently property 'm_iHealth' can be… missing?


sirherrbatka:
**Describe the bug**
I am not sure if this is bugged demo or what but this demo seems to indicate that there can be some sort of corner case during data tables parsing not covered.

**To Reproduce**
Check this demo file: https://antiochia.duckdns.org/index.php/s/zqf5KTq8PcmCFj9

**Expected behavior**
Not crash. 

**Library version**
beta.5

saul:
How old is this demo? It has a protocol version of 13942, whereas newer demos have 13964.

sirherrbatka:
It is on the older side, perhaps 2 weeks (at most) after CS2 went live.

markus-wa:
I guess it's fair to say we can just return 0 if the prop is missing :crossed_fingers:

akiver:
This demo comes from Faceit during the CS2 limited test phase.
The player prop `m_hPlayerPawn` doesn't point to a `CCSPlayerPawn` instance as it should but rather to a `CBaseModelEntity`.
That's why it crashes.

I did a test on this [branch](https://github.com/markus-wa/demoinfocs-golang/commit/dff72bcb36b0b6556a996eb4bd65dd13bf30525a), and even if we prevent the panic, we must patch all player's functions that rely on the pawn entity otherwise calling them would lead to a panic.

markus-wa:
Might not be worth it then to only support such a limited set of demos.

If anyone finds a nice fix I'd be happy to incorporate, but otherwise this might end up as a `wontfix`

jhaikonen:
I ran to the same issue with demo from esportal where the round has been resetted or something like that, and got the same error 'm_iHealth' missing. 

Ran the same demo with the branch mentioned by akiver (test-458) and it did not crash, but esportal demos don't have kill events etc unless you use the branch s2-fake-CMsgSource1LegacyGameEventList.

Anyway; I guess this is something to look into anyways and not just 'wontfix' 🤔

jhaikonen:
Update: merged test-458 to s2-fake-CMsgSource1LegacyGameEventList locally and was able to parse the demo ^^

markus-wa:
fixed in latest release - thanks @akiver

PR #489, opened 2023-12-29
fix: Participants().Playing() may returned disconnected players

### LINKED ISSUE #481: Player vanishes from connected players list, but continues to be active in the game.
**Describe the bug**
This demo https://antiochia.duckdns.org/index.php/s/ciywmyfkXe9jjxg displays curious behavior for player of steam id 76561198417781255 Namely after tick 20189 player is no longer on the connected player list, however, he also dies multiple times after this time mark, including tick 136067.
                                                                       
**To Reproduce**
https://antiochia.duckdns.org/index.php/s/ciywmyfkXe9jjxg (faceit demo)

**Expected behavior**
I reckon this is likely to be an issue with reconnected players handling.

**Library version**
v4.0.0

akiver:
It should fix #481, #458 and this [issue](https://github.com/akiver/cs-demo-manager/issues/723) (same problem as #458).

I noticed this thing with the demo from https://github.com/akiver/cs-demo-manager/issues/723:

disco

As you can see the last player is disconnected but still visible on the scoreboard.
When this kind of disconnection happens, the player pawn entity is "invalidated" - it points to a `CBaseViewModel` instance rather than a `CCSPlayerPawn` one.
This is why in #458 and https://github.com/akiver/cs-demo-manager/issues/723 the properties don't exist and it panics.

This PR:
- Update `player.IsConnected` when the pawn entity is invalid
- Verify that `m_hPawn` is valid, otherwise it means `m_hPlayerPawn` will point to a `CBaseViewModel`

Thanks

### DIFF
--- a/pkg/demoinfocs/common/player.go
+++ b/pkg/demoinfocs/common/player.go
@@ -37,6 +37,15 @@ type Player struct {
 }
 
 func (p *Player) PlayerPawnEntity() st.Entity {
+	pawn, exists := p.Entity.PropertyValue("m_hPawn")
+	if !exists {
+		return nil
+	}
+
+	if pawn.Handle() == constants.InvalidEntityHandleSource2 {
+		return nil
+	}
+
 	playerPawn, exists := p.Entity.PropertyValue("m_hPlayerPawn")
 	if !exists {
 		return nil
--- a/pkg/demoinfocs/common/player_test.go
+++ b/pkg/demoinfocs/common/player_test.go
@@ -431,6 +431,7 @@ func TestPlayer_Velocity(t *testing.T) {
 func createPlayerForVelocityTest() *Player {
 	controllerEntity := entityWithProperties([]fakeProp{
 		{propName: "m_hPlayerPawn", value: st.PropertyValue{Any: uint64(1), S2: true}},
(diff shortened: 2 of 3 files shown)


pkg/demoinfocs/common/player.go
```diff
@@ -37,6 +37,15 @@ type Player struct {
...
 func (p *Player) PlayerPawnEntity() st.Entity {
+	pawn, exists := p.Entity.PropertyValue("m_hPawn")
+	if !exists {
+		return nil
+	}
+
+	if pawn.Handle() == constants.InvalidEntityHandleSource2 {
+		return nil
+	}
+
```
akiver, line 48: An other way would be:

```go
func (p *Player) PlayerPawnEntity() st.Entity {
	playerPawn, exists := p.Entity.PropertyValue("m_hPlayerPawn")
	if !exists {
		return nil
	}

	entity := p.demoInfoProvider.FindEntityByHandle(playerPawn.Handle())
	if entity != nil && entity.ServerClass().Name() == "CCSPlayerPawn" {
		return entity
	}

	return nil
}
```

markus-wa (approved):

Issue #501, opened 2024-02-07
Parser crashes after CS2 update 2/6/2024


JuhaKiili:
**Describe the bug**
Parsing fails with `Fatal error: runtime: out of memory` for a demo after CS2 update 2/6/2024

**To Reproduce**
https://www.hltv.org/matches/2369635/sprout-vs-metizport-yalla-compass-spring-2024-contenders | Vertigo
https://www.hltv.org/matches/2369649/ecstatic-vs-tsm-european-pro-league-season-13 | Anubis

Code:
```go
package main

import (
	"fmt"
	"os"

	ex "github.com/markus-wa/demoinfocs-golang/v4/examples"
	demoinfocs "github.com/markus-wa/demoinfocs-golang/v4/pkg/demoinfocs"
	events "github.com/markus-wa/demoinfocs-golang/v4/pkg/demoinfocs/events"
)

// Run like this: go run test.go -demo /path/to/demo.dem
func main() {
	f, err := os.Open(ex.DemoPathFromArgs())
	checkError(err)
	defer f.Close()
	p := demoinfocs.NewParser(f)
	defer p.Close()

	p.RegisterEventHandler(func(e events.RoundEnd) {
		gs := p.GameState()
		fmt.Printf("Round %d parsed\n", gs.TotalRoundsPlayed())
	})

	err = p.ParseToEnd()
	checkError(err)
}

func checkError(err error) {
	if err != nil {
		panic(err)
	}
}
```
**Current behavior**
The parsing fails with `Fatal error: runtime: out of memory`

**Expected behavior**
The demos should parse without error and print out all the round numbers

**Library version**
Latest from master (e98a3fa02faa46032944acc23c3b0c0a6bf5be5d)

makindotcc:
I can say in my parser reader gets corrupted after reading entities, especially after reading ``m_flSimulationTime``. I think it is similar case here. Probably this field decoder format has been changed.

markus-wa:
See also https://github.com/markus-wa/demoinfocs-golang/issues/502 & https://github.com/markus-wa/demoinfocs-golang/issues/503

The underlying issue is likely the same - I will close the others in favour of this one

spheenik:
My hunch is that it has something to do with this:
```proto
message ProtoFlattenedSerializerField_t {
	message polymorphic_field_t {
		optional int32 polymorphic_field_serializer_name_sym = 1;
		optional int32 polymorphic_field_serializer_version = 2;
	}

[...]

PR #505, opened 2024-02-09
Support PVS bits

makindotcc:
Fixes issue https://github.com/markus-wa/demoinfocs-golang/issues/501
Explanation: https://github.com/markus-wa/demoinfocs-golang/issues/501#issuecomment-1935327804

I would be glad if someone test it if it properly works for older demos and any other cases.
Cs2 instead of continuing does similar thing to csgo:
https://github.com/perilouswithadollarsign/cstrike15_src/blob/f82112a2388b841d72cb62ca48ab1846dfcc11c8/engine/client.cpp#L1562-L1589

I think simple continue is fine :)


StefanDorresteijn:
I've tested this on post-update demos and a few pre-update demos and they all seem to work. Good work!

akiver:
Thank you ❤️

PR #535, opened 2024-05-02
feat: implement player_sound event

mateusmaia:
Literally what's written in the title.

Needed it for a little personal project, saw there was a "TODO" message for this event so assumed it's probably something others also want and here it is.

Not sure if there are any unit tests I should've updated, seems like there aren't but if there are I'll be happy to do it.

### DIFF
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -2,6 +2,7 @@ package demoinfocs
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/golang/geo/r3"
 	"github.com/markus-wa/go-unassert"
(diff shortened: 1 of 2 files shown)


pkg/demoinfocs/game_events.go
```diff
@@ -256,7 +257,7 @@ func newGameEventHandler(parser *parser, ignoreBombsiteIndexNotFound bool) gameE
 		"player_given_c4":                 nil,                                   // Dunno, only present in locally recorded (POV) demos
 		"player_ping":                     nil,                                   // When a player uses the "ping system" added with the operation Broken Fang, only present in locally recorded (POV) demos
 		"player_ping_stop":                nil,                                   // When a player's ping expired, only present in locally recorded (POV) demos
-		"player_sound":                    nil,                                   // When a player makes a sound. TODO: implement player_sound
+		"player_sound":                    delayIfNoPlayers(geh.playerSound),     // When a player makes a sound
```
mateusmaia, line 260: I'm not sure if this is the desired behavior, but I noticed that by using `delayIfNoPlayers` we lose the very first event even though it has a player.
This made me wonder if maybe this is happening to other events as well...

markus-wa: I think all the relevant events should hopefully already have `delayIfNoPlayers`

markus-wa (approved):

Issue #533, opened 2024-04-30
One of player has been written like unknown


Froctnow:
**Describe the bug**
There is demo, where one of player hasn't been triggered event PlayerConnected, but triggered PlayerDisconnected and he is wrote there like a unknown player
But if we investigate next events, we will find him there. For example event RoundEnd, we can find him in winners/losers and all okay

**To Reproduce**
Download link to an affected demo:  http://replay272.valve.net/730/003681822564448993333_0996605818.dem.bz2

**Expected behavior**
Player has been connected and written like normal player, and not unknown

**Library version**
v4.1.3

Markxy:
another more recent example: http://replay274.valve.net/730/003683095105276739793_0287984549.dem.bz2
with steamId `76561198327600344`

BestAwperEver:
There's a chance #537 will fix it. Could you try it and share the results?

Markxy:
> There's a chance #537 will fix it. Could you try it and share the results?

Worked for my case above 🙌

akiver:
Fixed with https://github.com/markus-wa/demoinfocs-golang/pull/537

PR #499, opened 2024-02-02
fix: possible wrong players team

akiver:
For some reason, the `team` field from `player_team` events may be incorrect.  
That's why players were assigned to the wrong team.

ref #494
ref #492

Issue #532, opened 2024-04-29
Some demos failing due to proto issue, on v4.1.3


DandrewsDev:
**Describe the bug**
During demo processing, demo processing crashes 
2024/04/28 22:36:21 failed to parse demo: failed to parse CMsgPlayerInfo msg: proto: cannot parse invalid wire-format data

**To Reproduce**
Download link to an affected demo: HLTV, ESEA, Matchmaking share link or other
https://www.faceit.com/en/csgo/room/1-1d796568-6470-439d-8606-2a630fc83aa3

Code:
```
func main() {
  f, err := os.Open("1-1d796568-6470-439d-8606-2a630fc83aa3-1-1.dem")
  if err != nil {
    log.Panic("failed to open demo file: ", err)
  }
  defer f.Close()

  p := demoinfocs.NewParser(f)
  defer p.Close()

  // Register handler on kill events
  p.RegisterEventHandler(func(e events.Kill) {
    var hs string
    if e.IsHeadshot {
      hs = " (HS)"
    }
    var wallBang string
    if e.PenetratedObjects > 0 {
      wallBang = " (WB)"
    }
    fmt.Printf("%s <%v%s%s> %s\n", e.Killer, e.Weapon, hs, wallBang, e.Victim)
  })

  // Parse to end
  err = p.ParseToEnd()
  if err != nil {
    log.Panic("failed to parse demo: ", err)
  }
}
```

**Expected behavior**
Demo processes, lists kill events.

**Library version**
v4 v4.1.3

**Additional context**
Full stack trace

```
...
Fre-e- <World> Fre-e-
ranch <World> ranch
SweatyMcRib <World> SweatyMcRib
2024/04/28 22:36:21 failed to parse demo: failed to parse CMsgPlayerInfo msg: proto: cannot parse invalid wire-format data
stacktrace:
goroutine 6 [running]:
runtime/debug.Stack()
        /opt/homebrew/opt/go/libexec/src/runtime/debug/stack.go:24 +0x64
github.com/markus-wa/demoinfocs-golang/v4/pkg/demoinfocs.NewParserWithConfig.func1({0x1009e5840, 0x14001ce3600})
        ~/go/pkg/mod/github.com/markus-wa/demoinfocs-golang/v4@v4.1.3/pkg/demoinfocs/parser.go:403 +0x2c
github.com/markus-wa/godispatch.(*Dispatcher).dispatchWithRecover.func1()
[...]

Issue #522, opened 2024-03-29
Possible panic since v4.1.0


akiver:
**Describe the bug**

Since the v4.1.0 that introduced POV support for CS2 demos, many fields may be `nil` and cause panic for consumers.

**To Reproduce**

Git clone the [CS demo analyzer repo](https://github.com/akiver/cs-demo-analyzer) and run the [tests](https://github.com/akiver/cs-demo-analyzer#tests) in verbose mode (`make test-verbose`).

You will have this [output](https://github.com/akiver/cs-demo-analyzer/pull/9/checks).
There is a panic and many warnings about fields being `nil` - it was not the case before v4.1.0.

**Expected behavior**

No panic or `nil` fields.

**Library version**
v4.1.0

**Additional context**

I didn't dig into it yet, maybe we should revert commits and release a v4.2.0 until it's resolved WDYT @markus-wa ?

markus-wa:
I think we should fix-forward on this. I do appreciate this can be a problem, but POV support is a valuable feature and we should keep supporting it while fixing all the problems with it.

PR #537, opened 2024-05-07
workaround for `userid` in `player_connect`

BestAwperEver:
Since the last update, I've been encountering many demos in which after a player has disconnected, and then reconnected, they wouldn't be registered in the `rawPlayers` field. Thus, in many cases, there were `nil` players in some events (e. g. all grenade events).

I investigated this a little and found out that seemingly the only place we could get information to add the player back to the `rawPlayers` (at least in particular demos I've been using) is inside the `player_connect` handler, which was disabled for s2 demos in #499. I'm assuming this was done because the user ids for players have been changed in the February, 6 update from being in the lower digits to 652xx.

I'm not sure what exactly has been changed in the last update (my guess is the `userinfo` string table is not parsed as often or not at all in some situations anymore?), but I've implemented a workaround that allows us to use the `player_connect` again by noticing that the old used ids are still present in the new ones -- they are the least significant byte of the two, and the other one is just set to 255.

Here's an [example demo](http://replay189.valve.net/730/003682163088783573210_2005656758.dem.bz2). On tick number 48430 `kefu_.` disconnects, then at tick 50731 we get a `player_connect` event, and then at tick 51237, when the `m_iConnected` is updated, `PlayerConnect` is dispatched (but, notably, we don't know the new user id at this point so the `rawPlayers` field doesn't get updated). This leads to `Thrower` being `nil` in the dispatched `HeExplode` event on tick 53754, which breaks at least my application 😄

This PR is more of an attempt to get some attention to the problem. There probably is a proper way to handle the situation without resorting to this hacky method, which might or might not work in the future. But I've already tested this implementation (and by "tested" I mean I've been using it in the production since yesterday) and was able to lower the error count in the parsed matches tenfold, so maybe that would help someone else in a similar situation.

### DIFF
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -641,10 +641,6 @@ func (geh gameEventHandler) HostageRescuedAll(map[string]*msg.CSVCMsg_GameEventK
 }
 
 func (geh gameEventHandler) playerConnect(data map[string]*msg.CSVCMsg_GameEventKeyT) {
-	if geh.parser.isSource2() {
-		return
-	}
-
 	pl := common.PlayerInfo{
 		UserID:       int(data["userid"].GetValShort()),
 		Name:         data["name"].GetValString(),
@@ -662,7 +658,14 @@ func (geh gameEventHandler) playerConnect(data map[string]*msg.CSVCMsg_GameEvent
 		}
 	}
 
-	geh.parser.setRawPlayer(int(data["index"].GetValByte()), pl)
+	var playerIndex int
+	if geh.parser.isSource2() {
+		playerIndex = pl.UserID
+	} else {
+		playerIndex = int(data["index"].GetValByte())
+	}
+
+	geh.parser.setRawPlayer(playerIndex, pl)
 }
 
 func (geh gameEventHandler) playerDisconnect(data map[string]*msg.CSVCMsg_GameEventKeyT) {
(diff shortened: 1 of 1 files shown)


akiver:
I think the root issue is how we parse stringtables since the version 4.1.0.
It's the same problem reported in https://github.com/markus-wa/demoinfocs-golang/issues/522 where many events now contain `nil` players.

I tried your demo against this branch https://github.com/markus-wa/demoinfocs-golang/tree/csdm which is v4.0.5 + updated proto messages and the thrower at tick `51237` is not `nil`.

If that can help it seems that the diff related to this issue is [here](https://github.com/markus-wa/demoinfocs-golang/pull/521/files#diff-b35f3237a3ea40004f7a0f982d5b4e45410ed4ea4d86ae617fb3b1e021d0aa34L378-L380).
We used to read the bit that indicates if there is a value even if there is no key but since v4.1.0 (for POV support) we read the value bit only if there is a key. Reverting this changes fix the problem.

akiver:
Just noticed that the change described in my previous comment may cause corrupted `CMsgPlayerInfo` messages as seen in https://github.com/markus-wa/demoinfocs-golang/issues/532 FYI @markus-wa

markus-wa:
@akiver would you be able to propose an MR? I'm not really sure what the best approach is since a few people rely on the POV features right now

akiver:
PR opened https://github.com/markus-wa/demoinfocs-golang/pull/540

akiver:
It looks like this PR is still required to fix https://github.com/markus-wa/demoinfocs-golang/issues/533 as this issue is not related to stringtables parsing.
Thanks @BestAwperEver!

akiver (approved):

Issue #539, opened 2024-05-16
All players are null for all matches of CBCS #4 tournament (Gamersclub)


JuhaKiili:
**Describe the bug**
All players are nil when parsing the demo.
All matches from CBCS4 tournament (Gamersclub servers) are the same
Demos work on some tools (Skybox) and CS2 built-in demo player

**To Reproduce**
[cbcs4-noplayers-1.dem](https://broken-demos-cs2lens.s3.eu-west-1.amazonaws.com/cbcs4-noplayers-1.dem)
[cbcs4-noplayers-2.dem](https://broken-demos-cs2lens.s3.eu-west-1.amazonaws.com/cbcs4-noplayers-2.dem)

Code:
`/examples/print-events/print_events.go`

**Current behaviour**
```
$ go run print_events.go -demo /home/juha/ws/demoinfocs-golang/cbcs4-noplayers-1.dem
? <M4A1 (HS)> ?
? <M4A1 (HS)> ?
? <M4A1 (HS)> ?
? <M4A1 (HS)> ?
? <M4A1 (HS)> ?
? <USP-S (HS)> ?
? <USP-S (HS)> ?
...
```

**Expected behavior**
```
$ go run print_events.go -demo /home/juha/ws/demoinfocs-golang/cbcs4-noplayers-1.dem
Player123 <M4A1 (HS)> Foobar
Lorem Ipsum <M4A1 (HS)> JohnDoe
HerpDerp <USP-S (HS)> Player123
JohnDoe <USP-S (HS)> HerpDerp
Player123 <M4A1 (HS)> Foobar
Lorem Ipsum <M4A1 (HS)> JohnDoe
HerpDerp <USP-S (HS)> Player123
...
```

**Library version**
Latest master (a58c851c08a21260cab660e329f96e27c366e3d2)

JuhaKiili:
> Could you try with this branch? #540

Works! Thank you!

Issue #525, opened 2024-04-24
Faceit - unexpected EOF: demo stream ended unexpectedly


maddenbjames:
**Describe the bug**
After parsing the entire demo, at the end of the parse it will show this error
failed to parse demo: unexpected EOF: demo stream ended unexpectedly (ErrUnexpectedEndOfDemo)

This has been happening a lot more as of late, and as far as I can tell almost exclusively with faceit demos.

**To Reproduce**
Demo page download
https://www.faceit.com/en/cs2/room/1-2304a144-2f75-4ab1-913d-2826bb857ae3

Run it on any parser, tested with our intricate one and also just a basic kill printer

Code:
```
func main() {
	f, err := os.Open("../demos/test.dem")
	if err != nil {
		log.Panic("failed to open demo file: ", err)
	}
	defer f.Close()

	p := dem.NewParser(f)
	defer p.Close()

	p.RegisterEventHandler(func(e events.Kill) {
		var hs string
		if e.IsHeadshot {
			hs = " (HS)"
		}
		var wallBang string
		if e.PenetratedObjects > 0 {
			wallBang = " (WB)"
		}
		fmt.Printf("%s <%v%s%s> %s\n", e.Killer, e.Weapon, hs, wallBang, e.Victim)
	})

	// Parse to end
	err = p.ParseToEnd()
	if err != nil {
		log.Panic("failed to parse demo: ", err)
	}
}
```

**Expected behavior**
It finishes the parse with no error. This error seems to happen either at the start of the demo or at the end of the demo parse.

**Library version**
v4.1.2

PR #540, opened 2024-05-16
fix: incomplete stringtables parsing

akiver:
fix https://github.com/markus-wa/demoinfocs-golang/issues/539 https://github.com/markus-wa/demoinfocs-golang/issues/532 https://github.com/markus-wa/demoinfocs-golang/issues/522 and maybe https://github.com/markus-wa/demoinfocs-golang/issues/525 (i can't reproduce the issue).
Tests of the project https://github.com/akiver/cs-demo-analyzer/pull/10/checks are also green.

This PR reverts the logic in `parseStringTable` introduced in v4.1.0 and instead if a panic occurs while parsing a stringtable, we recover and dispatch a warning.
I'm not 100% sure but sometimes data seems corrupted and skipping it seems fine.
I didn't find a logic that doesn't panic when parsing stringtables for both CSTV and POV demos, that's why I decided to go for the recover solution.

### DIFF
--- a/pkg/demoinfocs/stringtables.go
+++ b/pkg/demoinfocs/stringtables.go
@@ -294,7 +294,9 @@ const (
 )
 
 // Parse a string table data blob, returning a list of item updates.
-func parseStringTable(
+//
+//nolint:funlen,gocognit
+func (p *parser) parseStringTable(
 	buf []byte,
 	numUpdates int32,
 	name string,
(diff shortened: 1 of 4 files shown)


pkg/demoinfocs/stringtables.go
```diff
@@ -308,6 +310,16 @@ func parseStringTable(
 		return items
 	}
 
+	defer func() {
+		err := recover()
+		if err != nil {
+			p.eventDispatcher.Dispatch(events.ParserWarn{
```
markus-wa, line 316: could we re-try parsing with the previous logic here? or is there no point?

JuhaKiili:
I have tested with couple of POV demos too, and this branch works ok 👍

pkg/demoinfocs/stringtables.go
akiver: it always leads to empty items

PR #544, opened 2024-06-11
fix: possible nil player in game events

akiver:
fix https://github.com/akiver/cs-demo-manager/issues/863


examples/map_metadata_test.go
```diff
@@ -9,6 +9,8 @@ import (
 )
 
 func TestGetMapMetadata(t *testing.T) {
+	t.Skip()
```
akiver, line 12: @markus-wa I skip tests that rely on https://radar-overviews.csgo.saiko.tech because it's currently down

PR #561, opened 2024-08-08
Add Kukri's knife to equipment map

jouderianjr:
The Kukri knife was added sometime ago and because of that I'm getting a warning whenever a player is using it in the demo:

`unknown equipment with index 526`

526 is the id for that knife in the items file


akiver:
Thanks!

PR #592, opened 2025-06-05
feat: Add ability to retrieve player viewmodel settings

captainswain:
This pr introduces functionality to extract player viewmodel settings (offsets and field of view)

### Changes
* Added `ViewmodelOffset` and `ViewmodelFOV` methods to the `Player` struct to retrieve viewmodel settings. These methods return zero values for CS:GO demos or when player data is unavailable.
* Added relevant tests
* Added example + readme


markus-wa:
nice, thx!

PR #616, opened 2025-11-26
Gracefully handle invalid Content-Type header from CSTV broadcasts

cadenkriese:
The Starladder Budapest Majors' CSTV feeds redirect all invalid requests to the major's home webpage. Since those redirects end up returning 200, demoinfocs will append the page HTML to the buffer, causing a panic. My solution is to treat responses that don't use Content-Type: application/octet-stream as errors.

### DIFF
--- a/pkg/demoinfocs/cstv/cstv.go
+++ b/pkg/demoinfocs/cstv/cstv.go
@@ -51,7 +51,7 @@ func (c *Reader) Read(p []byte) (n int, err error) {
 
 		defer deltaResp.Body.Close()
 
-		if deltaResp.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


markus-wa:
Sorry for the slow reply

If there is a reliable way to detect it we can do that, but we need to make sure there is no chance we won't be able to parse legit broadcasts.

We could add an inverse content-type detection. e.g. if it's `text/html` we can be reasonably sure it's not valid.

At first I thought we could avoid redirects but there is legit cases where we would want to follow a redirect (e.g. if fragments are hosted on a public bucket and the api redirects to the bucket)

markus-wa:
I checked - CS2 accepts non-octet-stream content-types, it seems to ignore it, so unfortunately we can't make this change

cadenkriese:
Fair enough. Thank you for checking. Hopefully future Major organizers won't have such odd behavior with their CDN.

PR #627, opened 2026-01-05
feat: add fields/functions to get player buttons related info

akiver:
Adds fields and functions to get information about buttons currently pressed by players.
It also adds a `PlayerButtonsStateUpdate` event dispatched when a player pressed/released buttons.

- `player#ButtonsPressedState` to get the raw uint64 state
- `player#IsPressingButton(btnMask)` to check if the player is pressing a  specific button
- `player#InspectWeaponCount` number of times the player has inspected his weapon

It would allow us to generate stats like this https://blog.scope.gg/inspect-weapon-stats-beijing-en/


pkg/demoinfocs/sendtables/sendtablescs2/entity.go
```diff
@@ -161,8 +161,6 @@ func (e *Entity) PropertyValueMust(name string) st.PropertyValue {
 }
 
 const (
-	serverClassPlayer = "CCSPlayerPawn"
```
akiver: deadcode

markus-wa (approved):

PR #636, opened 2026-02-10
fix: Reset CSTV delta failure count on success

mrunelov:
Ran into this issue where 404s to /delta accumulates and triggers EOF even though the polling eventually succeeds.

I don't have much experience with CSTV broadcast servers or other polling internals in this lib, but at least the one I'm testing with is consistently behind the library polling, so `nFails` always reaches its limit in less than a minute.

### DIFF
--- a/pkg/demoinfocs/cstv/cstv.go
+++ b/pkg/demoinfocs/cstv/cstv.go
@@ -74,7 +74,8 @@ func (c *Reader) Read(p []byte) (n int, err error) {
 		backoff = time.Second
 		nFails = 0
 
-		n2, err := c.buf.Read(p[n:])
+		var n2 int
+		n2, err = c.buf.Read(p[n:])
 		n += n2
 	}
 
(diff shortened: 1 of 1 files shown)


mrunelov:
I think I might've misunderstood the way the buffer is filled here, there are bugs, but I will investigate more, I don't think this is the proper fix.

markus-wa:
Let me know if there's anything I can help with regarding the investigation!

mrunelov:
> Let me know if there's anything I can help with regarding the investigation!

@markus-wa Thank you! I'm making good progress with the cstv loop, it seems to be working now (locally).

However, I had to do an unrelated change: I kept getting `unable to find existing entity`, which seems to be benign, but it triggers panics and crashes the parser. Doing `continue` there instead locally. Wouldn't mind your thoughts on that as I don't know much about the internals, maybe it's dropping data and then hitting this. I can do a PR for my CSTV fixes soon.

Edit: I think both problems I've seen is due to bugs in the server implementation. Been testing the reference impl and this one: https://github.com/FlowingSPDG/gotv-plus-go, and the second one seems to return 200 OK with empty responses sometimes (this causes the 404s as frag is incremented), and also somehow cause the missing entities, so I guess it's dropping data somewhere along the way. The reference implementation does not have these issues.

PR #639, opened 2026-02-17
Add csskill.com to project list in README

csskill:
Added csskill.com as a CS2 performance and analytics platform.


csskill:
Licensing should now be compatible under https://csskill.com/licenses

PR #644, opened 2026-04-04
perf: reduce CPU and allocation overhead in CS2 entity parsing

markus-wa:
- Replace map[int]*fpNameTreeCache with []*fpNameTreeCache for O(1) slice indexing instead of hash map lookups, eliminating ~10s of mapaccess overhead
- Skip getNameForFieldPath in readFields when no update handlers are registered, avoiding millions of string lookups in the common case
- Reuse strBuf across readString calls in reader to avoid per-call allocation
- Increase fieldState initial capacity from 8 to 64 to reduce slice reallocations for entities with many fields

Result: ~12-15% throughput improvement on the demo parsing benchmarks.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,12 @@ exclusions:
     - '*\\.pb\\.go$'
     - parser_interface.go
 
+issues:
+  exclude-rules:
+    - path: "pkg/demoinfocs/sendtables/sendtablescs2"
+      linters:
+        - godox
+
 formatters:
   enable:
     - gci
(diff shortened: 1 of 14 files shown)


markus-wa:
@copilot add nolint for old issues and issues about cognitive complexity aince the forced inlining/deep nesting is done on purpose for perf reasons

markus-wa:
@copilot remove dead code, dont use nolint for those

markus-wa:
@copilot fix linting issues

Issue #406, opened 2023-08-31
[CS2] Player's velocity / eye position are missing


akiver:
**Describe the bug**

The player's velocity and the eye position (z) are unavailable with CS2 demos.

**Expected behavior**

Velocity and eye position should be correct.

**Library version**
Branch `cs2`

**Additional context**

The props used to retrieve these data with Source 1 demos are not present with CS2 demos and it seems that there is no equivalent.

markus-wa:
### velocity

v5 will remove velocity, but we can bring it back as an optional setting in the future where we calculate it using the difference between current and last position

### eye pos

if we have a reliable way of getting eye / camera position I would love to implement this  in the parser

markus-wa:
closing as out of scope at this time

PR #647, opened 2026-04-12
Add player eye-position helper for CS2 pawns

WangChuDi:
## Related issue
- relates to #406

## Summary
- add `common.Player.PositionEyes()` for CS2 pawns by combining `Position()` with pawn plain `m_vecX/m_vecY/m_vecZ`
- make `common.Player.PositionEyes()` return `(r3.Vector, bool)` so callers can distinguish true eye-position data from fallback coordinates
- keep the fix scoped to ViewOffset functionality without changing `sendtablescs2` serializer naming
- preserve the existing world-position path based on `CBodyComponent` while exposing a higher-level eye-position helper

## Authorship
This pull request was written with assistance from **gpt-5.4**.

## Why this approach
This PR intentionally fixes only the eye-position functionality.

In the current CS2 path, the generic semantic property path for ViewOffset is not reliably exposed without deeper serializer work, but real-demo probing showed that plain pawn `m_vecX/m_vecY/m_vecZ` consistently behave like eye-offset components on `CCSPlayerPawn`.

That makes `Player.PositionEyes()` a small, functional fix for callers that need eye position, without changing the generic property-resolution contract.

The `(r3.Vector, bool)` return shape keeps the data fidelity explicit:
- `PositionEyes()` returns `(position, true)` only when the eye-offset fields were available
- if the pawn exists but the eye-offset fields are unavailable, it returns the pawn base position and `false`
- if there is no pawn, it returns the zero vector and `false`

## Real-demo evidence
Local probing was done against two real CS2 demos:
- `cs-demos/cs2/esportal_6008132_2023_mirage.dem`
- `cs-demos/cs2/challengermode_6c306e56-8170-4092-b402-08dbf813e452.dem`

Observed on `CCSPlayerPawn`:
- `m_vecViewOffset` -> missing
- `CBaseModelEntity.m_vecViewOffset` -> missing
- plain `m_vecX/m_vecY/m_vecZ` -> present
- `CBodyComponent.m_vecX/m_vecY/m_vecZ` -> also present, but clearly part of world-position reconstruction rather than eye offset

### Example samples
From real demo probing:

```text
frame=16 tick=1   offset=(0.00, 0.00, 64.00)
frame=33 tick=17  offset=(0.00, 0.00, 64.68)
frame=42 tick=26  offset=(0.31, 0.56, 54.35)
frame=43 tick=27  offset=(0.00, 0.00, 46.68)
frame=661 tick=545 offset=(0.00, 0.00, 64.00)
frame=662 tick=546 offset=(0.00, 0.00, 61.00)
frame=801 tick=683 offset=(0.00, 0.00, 61.12)
```

These values are plausible as eye-height offsets and not plausible as body/world-position offsets.

## Cross-check against crouch signals
I also compared plain pawn `m_vecZ` against crouch-related signals already exposed by `common.Player`:
- `IsDucking()`
- `IsDuckingInProgress()`
- `IsUnDuckingInProgress()`
- `IsPressingButton(common.ButtonDuck)`
- pawn `m_pMovementServices.m_flDuckAmount`
- pawn `m_pMovementServices.m_bDesiresDuck`
- pawn `m_fFlags`

Observed same-frame correlation from a real demo:

```text
frame=16 tick=1
  offset.Z=64.00 ducking=false duck_button=false

frame=33 tick=17
  offset.Z=64.68 ducking=false duck_button=false

frame=42 tick=26
  offset.Z=54.35 ducking=true duck_button=true

frame=43 tick=27
  offset.Z=46.68 ducking=true duck_button=true

frame=50 tick=34
  offset.Z=60.27 ducking=false duck_button=false

frame=51 tick=35
  offset.Z=64.68 ducking=false duck_button=false
```

### Full duck / unduck transition example
From `cs-demos/cs2/esportal_6008132_2023_mirage.dem` for player `AsSeLeLe`:

| stage | frame | tick | offset.Z | ducking | ducking_progress | unducking_progress | duck_button |
|---|---:|---:|---:|---|---|---|---|
| standing before | 16 | 1 | 64.00 | false | false | false | false |
| ducking in progress | 1526 | 1370 | 63.86 | false | true | false | true |
| fully ducked | 1539 | 1383 | 46.00 | true | false | false | true |
| unducking in progress | 1613 | 1452 | 52.01 | false | false | true | false |
| standing after | 1619 | 1458 | 64.00 | false | false | false | false |

This sequence shows the expected state-machine behavior directly from a real CS2 demo: `ducking_progress` tracks the transition into crouch, `ducking` marks the fully crouched state, and `unducking_progress` tracks the transition back to standing while the eye offset rises again.

This is strong behavioral evidence that plain pawn `m_vecZ` is tracking eye height / ViewOffset behavior rather than body position.

## Why not use `CBodyComponent.m_vec*`
`CBodyComponent.m_cellX/Y/Z` and `CBodyComponent.m_vecX/Y/Z` are already used by `Entity.Position()` to reconstruct world position:

```go
Position = coordFromCell(CBodyComponent.m_cell*, CBodyComponent.m_vec*)
```

Probe values for `CBodyComponent.m_vecZ` looked like position-related offsets, for example:
- `811.3275`
- `1.5569626`

Those do not match plausible eye-height offsets, while plain pawn `m_vecZ` does.

## Scope and limitations
This PR does **not** claim to fully repair generic CS2 semantic property lookup for ViewOffset.
It only provides a higher-level helper for the concrete functionality needed by callers that want player eye position.

If generic `PropertyValue("m_vecViewOffset.m_vecZ")` support is desired later, that should be handled as a separate serializer/sendtables change.

## Verification
- `go build ./...`
- `go test ./pkg/demoinfocs/common`
- local real-demo probing on two CS2 demos
- local crouch-signal correlation showing `m_vecZ` drops during ducking and rises again when standing
- analyzer-side verification with `go test ./pkg/api` and `go build ./...`


pkg/demoinfocs/common/player.go
```diff
@@ -419,6 +419,81 @@ func (p *Player) Position() r3.Vector {
...
+	pos := pawnEntity.Position()
+	offset, ok := eyePositionOffset(pawnEntity)
+	if !ok {
+		return pos, false
+	}
+
+	return pos.Add(offset), true
+}
+
+func eyePositionOffset(entity st.Entity) (r3.Vector, bool) {
```
markus-wa: can be a method of *Player rather than take entity arg

pkg/demoinfocs/common/player.go
```diff
@@ -419,6 +419,81 @@ func (p *Player) Position() r3.Vector {
...
+	}
+
+	return r3.Vector{
+		X: float64(x),
+		Y: float64(y),
+		Z: float64(z),
+	}, true
+}
+
+func floatProperty(entity st.Entity, propName string) (float32, bool) {
```
markus-wa: i think we have some other utility fucs - move this to the same file as the other utilites and make sure naming & behaviour is consistent

pkg/demoinfocs/common/player.go
```diff
@@ -419,6 +419,81 @@ func (p *Player) Position() r3.Vector {
...
 }
 
+// PositionEyes returns the best available in-game coordinates for the player's eye position.
+//
+// When a player pawn exists and CS2 eye-offset fields are available, it returns the pawn base
+// position plus the eye offset. When the pawn exists but the eye-offset fields are unavailable,
+// it falls back to the pawn base position. When there is no pawn, it returns the zero vector.
+//
+// See also PositionEyesOK().
+func (p *Player) PositionEyes() r3.Vector {
```
markus-wa: I think we should rename `PositionEyesOK` to `PositionEyes` and just not have a utility wrapper.

That way it's more obvious that this is not guaranteed to work - Users are much more likely to actually read the function docs if there are two returned values rather than just one

markus-wa:
thanks - looks good overall, just some minor comments

WangChuDi:
I’ve made some changes. Could you take a look and let me know if there are any issues? @markus-wa

markus-wa:
amazing, thanks!

Issue #642, opened 2026-03-30
CS2: when player_falldamage is missing, PlayerHurt fallback can misclassify fall/world damage as C4


WangChuDi:
**Describe the bug**
In CS2, when a real fall/world damage case does not emit a usable `player_falldamage` marker, typed `events.PlayerHurt` can incorrectly fall back to `EqBomb` / `C4`.

The key problem is the fallback path in `attackerWeaponType()` for `player_hurt weapon=""` / `EqUnknown`.

**To Reproduce**
Affected demo links:
- https://www.hltv.org/matches/2385959/faze-vs-inner-circle-esl-pro-league-season-22
- https://www.hltv.org/matches/2389634/bcgame-vs-ninjas-in-pyjamas-iem-krakw-2026-stage-1

Observed CS2 fall/world samples:

1. `broky` (`faze-vs-inner-circle-m2-nuke.dem`, round 21, tick 170355)
- generic `player_hurt.weapon = ""`
- generic `player_death.weapon = "worldent"`
- no `player_falldamage`
- typed `PlayerHurt.Weapon = C4`
- typed `Kill.Weapon = World`

2. `s1mple` (`bc-game-vs-ninjas-in-pyjamas-m2-nuke.dem`, round 4, tick 24826)
- generic `player_hurt.weapon = ""`
- generic `player_death.weapon = "worldent"`
- no `player_falldamage`
- typed `PlayerHurt.Weapon = C4`
- typed `Kill.Weapon = World`

The second sample is not a bomb-end round, so this is not just a bomb-end-only problem.

**Expected behavior**
If `player_hurt.weapon` is empty and there is no usable `player_falldamage` marker, typed `PlayerHurt.Weapon` should not fall through to `EqBomb/C4` for world/fall damage.

**Library version**
v5.1.2

**Additional context**
The important path seems to be:

```go
wepType := common.MapEquipment(data["weapon"].GetValString())
wepType = geh.attackerWeaponType(wepType, userID)
```

For these samples:
- `player_hurt.weapon = ""` -> `MapEquipment("") == EqUnknown`
- no usable `player_falldamage` marker -> no promotion to `EqWorld`
- then `attackerWeaponType()` falls through its unknown-weapon fallback and typed `PlayerHurt.Weapon` becomes `EqBomb/C4`

By contrast, typed `Kill` still works because `player_death.weapon = "worldent"` maps directly to `EqWorld`.

So the core issue seems to be:

[...]

PR #648, opened 2026-04-15
Fix empty-weapon PlayerHurt world vs bomb classification

WangChuDi:
## Summary
- delay only ambiguous CS2 `player_hurt` events with an empty weapon string until end-of-frame so world/fall damage can be distinguished from same-frame C4 damage
- keep real bomb damage classified as bomb by correlating both game-event and datatable `BombExplode` paths, while avoiding the old zero-value round-end fallback
- keep the final implementation scoped to the ambiguous empty-weapon path rather than changing dispatch timing for all `PlayerHurt` events

## Validation
- `go test ./pkg/demoinfocs -run 'Test(AttackerWeaponType|PlayerHurt)_'`
- real-demo check on `test/cs-demos/cs2/faze-vs-inner-circle-m2-nuke.dem`

## Comparison to issue #642
Issue #642 reports that some CS2 world/fall samples emit raw `player_hurt.weapon = ""`, but typed `PlayerHurt.Weapon` incorrectly falls through to `C4`, while typed `Kill.Weapon` still ends up as `World`.

### Issue-reported samples (before this fix)
From issue #642:

| sample | demo | round / tick | raw `player_hurt.weapon` | usable `player_falldamage` | raw `player_death.weapon` | typed `PlayerHurt.Weapon` | typed `Kill.Weapon` |
|---|---|---|---|---|---|---|---|
| `broky` world death | `faze-vs-inner-circle-m2-nuke.dem` | round 21 / tick 170355 | `""` | no | `"worldent"` | `C4` | `World` |
| `s1mple` world death | `bc-game-vs-ninjas-in-pyjamas-m2-nuke.dem` | round 4 / tick 24826 | `""` | no | `"worldent"` | `C4` | `World` |

These are the mismatches described in the issue: the generic `player_hurt` path falls through to `C4`, while the generic `player_death` path still maps to `World`.

### Real-demo world / fall samples after this fix
From `test/cs-demos/cs2/faze-vs-inner-circle-m2-nuke.dem`:

| sample | frame | tick | raw `player_hurt.weapon` | same-frame `BombExplode` | raw `player_death.weapon` | typed `PlayerHurt.Weapon` | typed `Kill.Weapon` |
|---|---:|---:|---|---|---|---|---|
| non-lethal world sample (`Twistzz`) | 2526 | 2323 | `""` | no | – | `World` | – |
| non-lethal world sample (`Flierax`) | 21416 | 20011 | `""` | no | – | `World` | – |
| issue-style world death (`broky`) | 181991 | 170355 | `""` | no | `"worldent"` | `World` | `World` |

This is the behavior the issue asks for: the world/fall samples no longer collapse into `C4`, and the `broky` sample now matches `PlayerHurt=World` with `Kill=World` instead of the mismatch described in the issue.

### Same-frame bomb samples after this fix
From the same real demo:

| sample | frame | tick | raw `player_hurt.weapon` | same-frame `BombExplode` | typed `PlayerHurt.Weapon` |
|---|---:|---:|---|---|---|
| bomb damage sample 1 | 14132 | 13196 | `""` | yes | `C4` |
| bomb damage sample 2 | 43767 | 40941 | `""` | yes | `C4` |
| bomb damage sample 3 | 78816 | 73759 | `""` | yes | `C4` |
| bomb damage sample 4 | 141156 | 132116 | `""` | yes | `C4` |
| bomb damage sample 5 | 160681 | 150401 | `""` | yes | `C4` |

So the real-demo result now matches the intended split described in the issue comments: empty-weapon world/fall samples stay on the `World` side, while true same-frame bomb damage stays on the `C4` side.

### DIFF
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -1065,8 +1065,6 @@ func (geh gameEventHandler) attackerWeaponType(wepType common.EquipmentType, vic
 		}
 	}
 
-	unassert.NotSame(wepType, common.EqUnknown)
-
 	return wepType
 }
 
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -1056,6 +1056,7 @@ func (geh gameEventHandler) attackerWeaponType(wepType common.EquipmentType, vic
 	// unfortunately RoundEndReasonTargetBombed isn't enough and sometimes we need to check for 0 as well
 	if wepType == common.EqUnknown {
 		if reason, ok := geh.frameToRoundEndReason[geh.parser.currentFrame]; ok {
+			//nolint:exhaustive
 			switch reason {
 			case 0:
 				fallthrough
--- a/pkg/demoinfocs/game_events.go
+++ b/pkg/demoinfocs/game_events.go
@@ -486,47 +486,29 @@ func (geh gameEventHandler) playerHurt(data map[string]*msg.CMsgSource1LegacyGam
 		})
 	}
 
-	// CS2 may emit player_hurt with an empty weapon string for both world/fall damage and C4 damage.
-	// Delay only these ambiguous cases until the end of the frame so same-frame fall-damage, BombExplode,
-	// and round-end evidence is available without changing dispatch timing for known weapons.
 	if rawWeapon == "" && wepType == common.EqUnknown {
 		geh.parser.delayedEventHandlers = append(geh.parser.delayedEventHandlers, func() {
-			dispatchPlayerHurt(geh.playerHurtWeaponType(wepType, userID))
+			resolvedType := geh.attackerWeaponType(wepType, userID)
+			if resolvedType == common.EqUnknown {
+				if geh.frameToBombExploded[geh.parser.currentFrame] {
+					resolvedType = common.EqBomb
+				} else {
+					resolvedType = common.EqWorld
+				}
+			}
+
+			dispatchPlayerHurt(resolvedType)
 		})
 
 		return
 	}
 
-	dispatchPlayerHurt(geh.attackerWeaponType(wepType, userID))
+	wepType = geh.attackerWeaponType(wepType, userID)
+	dispatchPlayerHurt(wepType)
 }
 
 func (geh gameEventHandler) playerFallDamage(data map[string]*msg.CMsgSource1LegacyGameEventKeyT) {
 	geh.userIDToFallDamageFrame[data["userid"].GetValShort()] = geh.parser.currentFrame
-}
-
-func (geh gameEventHandler) playerHurtWeaponType(wepType common.EquipmentType, victimUserID int32) common.EquipmentType {
-	if wepType != common.EqUnknown {
-		return geh.attackerWeaponType(wepType, victimUserID)
-	}
-
-	if geh.userIDToFallDamageFrame[victimUserID] == geh.parser.currentFrame {
-		return common.EqWorld
-	}
-
-	if geh.frameToBombExploded[geh.parser.currentFrame] {
-		return common.EqBomb
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/demoinfocs/game_events.go
```diff
@@ -1029,16 +1055,16 @@ func (geh gameEventHandler) attackerWeaponType(wepType common.EquipmentType, vic
...
 	// unfortunately RoundEndReasonTargetBombed isn't enough and sometimes we need to check for 0 as well
 	if wepType == common.EqUnknown {
-		switch geh.frameToRoundEndReason[geh.parser.currentFrame] {
-		case 0:
-			fallthrough
-		case events.RoundEndReasonTargetBombed:
-			wepType = common.EqBomb
+		if reason, ok := geh.frameToRoundEndReason[geh.parser.currentFrame]; ok {
+			switch reason {
+			case 0:
```
markus-wa: what's 0? are we missing a constant or why is this required?

pkg/demoinfocs/game_events.go
```diff
@@ -1029,16 +1055,16 @@ func (geh gameEventHandler) attackerWeaponType(wepType common.EquipmentType, vic
 	// if the round ended in the current frame with reason 1 or 0 we assume it was bomb damage
 	// unfortunately RoundEndReasonTargetBombed isn't enough and sometimes we need to check for 0 as well
 	if wepType == common.EqUnknown {
-		switch geh.frameToRoundEndReason[geh.parser.currentFrame] {
-		case 0:
-			fallthrough
-		case events.RoundEndReasonTargetBombed:
-			wepType = common.EqBomb
+		if reason, ok := geh.frameToRoundEndReason[geh.parser.currentFrame]; ok {
+			switch reason {
```
markus-wa:
```suggestion
			//nolint:exhaustive
			switch reason {
```

markus-wa:
LGTM, one comment

pkg/demoinfocs/game_events.go
```diff
@@ -1052,13 +1052,14 @@ func (geh gameEventHandler) attackerWeaponType(wepType common.EquipmentType, vic
 		wepType = common.EqWorld
 	}
 
-	// if the round ended in the current frame with reason 1 or 0 we assume it was bomb damage
-	// unfortunately RoundEndReasonTargetBombed isn't enough and sometimes we need to check for 0 as well
+	// For this unknown-weapon fallback, real same-frame bomb samples are not always covered by
+	// RoundEndReasonTargetBombed alone. Some of them still expose RoundEndReasonStillInProgress
+	// on the same frame, so we also treat that value as bomb-side evidence here.
```
markus-wa: hmm - this sounds like it will still not always be accurate, right?

essentially this will only fix the issue if it happens after the round has ended, not before, correct?

WangChuDi: I originally considered RoundEndReasonStillInProgress because the parser already documents that the bomb props used to detect bomb events are updated after m_eRoundWinReason, so it seemed plausible that some same-frame bomb cases might still expose an in-progress round-end value before the round-end state fully settles.

After re-checking the real issue demos, I did not actually observe round_end.reason == RoundEndReasonStillInProgress in those reproduced bomb-damage frames.

Perhaps we should keep the “RoundEndReasonStillInProgress” case. Would it be okay to mention this in the comments?
`As the bomb props used to detect bomb events are updated after m_eRoundWinReason, the case of RoundEndReasonStillInProgress is retained`

markus-wa: looking at this logic again I think the old documentation + logic was completely off

WangChuDi: I rescanned the 20 demos in test/cs-demos/cs2, and none of these demos showed RoundEndReasonStillInProgress.

markus-wa: yeah - I removed that stuff now

markus-wa:
LGTM

PR #659, opened 2026-07-07
fix: always update player user ID in getOrCreatePlayer

akiver:
ref https://github.com/akiver/cs-demo-analyzer/pull/28


akiver:
FYI @markus-wa

PR #661, opened 2026-07-17
fix: PlayerButtonsStateUpdate not working since last CS2 update

akiver:
The prop `m_nButtonDownMaskPrev`, used to dispatch `PlayerButtonsStateUpdate` events, has been removed in the 09/07/2026 update.
A workaround is to read user commands from `CSVCMsg_UserCommands` messages.
The downside is that it slows parsing since we have to unmarshal many more messages than before (from ~5s to ~7s with this PR and 1 specific demo).

@markus-wa what do you think? We could also add an option to enable internal user commands parsing?


akiver:
@markus-wa another option would be to parse the value of `Buttonstate1` (it corresponds to the old prop `m_nButtonDownMaskPrev`) without `proto.Unmarshal` and any allocation. I did in the v4 branch https://github.com/markus-wa/demoinfocs-golang/commit/599c25ed221f6573a57e55f3411c99635b088b7e and there is _almost_ no performance overhead.
We could also add a full example on how to read and parse `CSVCMsg_UserCommands` messages.

markus-wa:
hmm, what other data is in these user commands?

akiver:
> hmm, what other data is in these user commands?

Positions, view angles, history of inputs...
See https://github.com/SteamTracking/GameTracking-CS2/blob/master/Protobufs/cs_usercmd.proto#L39 and https://github.com/SteamTracking/GameTracking-CS2/blob/master/Protobufs/usercmd.proto#L30

markus-wa:
Let's add a config option to explicitly disable this parsing - but default to parsing everything

markus-wa:
merged as part of #669

Issue #156, opened 2019-11-07
Kill.Killer may be nil


Linus4:
**Describe the bug**
The Killer pointer in the events.Kill type may be nil. In my case, the player fell to his death.
Similar to #151 .

**To Reproduce**
[Link to demo](https://user.informatik.uni-goettingen.de/~linus.weber/s11_severity_de_inferno.dem)

Code:
[Line in parser code](https://github.com/markus-wa/demoinfocs-golang/blob/d7347829b857e430b495788927511915076081ea/game_events.go#L279)

```go
type Kill struct {
	KillerName string
	KillerTeam common.Team
	VictimName string
	VictimTeam common.Team
	Weapon     string
}

parser.RegisterEventHandler(func(e event.Kill) {	
	kill := ocom.Kill{
		KillerName: e.Killer.Name,
		KillerTeam: e.Killer.Team,
		VictimName: e.Victim.Name,
		VictimTeam: e.Victim.Team,
		Weapon:     e.Weapon.Weapon.String(),
}
```

**Expected behavior**
`Kill.Killer` should be the same as `Kill.Victim`. Also, the `Kill.Weapon` could be something like suicide or falling instead of `UNKNOWN`. Maybe don't dispatch a Kill event at all or add a comment in the documentation.

**Library version**
v1.3.3

Linus4:
I just discovered that the Victim pointer in the Kill type may also be nil. It occurred when a player left the server, presumably due to his game crashing.

markus-wa:
Hey @Linus4 

`Killer` being `nil` is intended. IIRC the issue is that we cannot differentiate between different kinds of `UNKNOWN`. It could be self damage from a molotov, falling or the `kill` command.

I will update the docs accordingly.

> I just discovered that the Victim pointer in the Kill type may also be nil. It occurred when a player left the server, presumably due to his game crashing.

Interesting, I will try to look into this.

Linus4:
I also just found that the Shooter from the WeaponFire event may be nil. All shot from one player had the nil pointer until he reconnected to the server.

markus-wa:
I looked into `Kill.Killer` a bit more and noticed that in this case for the demo you linked we should have returned `EqWorld` as weapon since the event has the `weapon` attribute as `weapon_worldspawn`.

In addition I'm adding the comment `May be nil for world damage (EqWorld)` to `Kill.Killer`

[...]

PR #663, opened 2026-07-27
Various fixes: correctness, new events/getters, and doc corrections

markus-wa:
## Summary

A batch of correctness fixes, new events/getters, and doc corrections, driven by an investigation against 13 recent CS2 **broadcast-GOTV** demos (BLAST Bounty 2026 S2, post-Mar-2026 patch). Each item is a self-contained commit and was verified by running code against real demos; the golden fixture was regenerated for the two behavior changes it captures.

`go build`, `go vet` and `go test ./pkg/...` all pass.

## Bug fixes

- **Cap fatal-hit `PlayerHurt.HealthDamageTaken` at remaining HP.** On a fatal hit the handler used `player.Health()`, which reads start-of-tick HP (entity state is applied after the tick's events dispatch). For same-tick multi-attacker kills this over-reported the finishing blow by the earlier attacker's damage. Now tracks cumulative health damage per victim per tick. *Verified: genuine inflations (>1) dropped to 0, benign +1 rounding preserved.*
- **Dispatch `OtherDeath` for CS2 `entity_killed`.** The CS2 `entity_killed` game-event was registered with a no-op handler, so `events.OtherDeath` never fired despite ~180 `entity_killed`/match. Added a CS2 handler that resolves the entity-index payload, skips player-pawn deaths (already covered by `player_death`), and emits `OtherDeath` for non-player entities (chickens, props, doors, …). *Verified: 0 → ~34 events/demo (CDynamicProp etc.).*
- **`PlayerFlashed` victim pairing by detonation, not FIFO.** Victims and detonations were both written to the head of a FIFO queue of in-flight flashbangs; a zero-victim detonation while another flashbang was airborne shifted attribution by one for the rest of the round. Now keys the detonated entry by the destroyed projectile's identity, buffers pawns flashed each frame, and attributes them to the flashbang that detonated that frame (closest by detonation position when several detonate at once). The thrower (`proj.Thrower`) was already correct, matching `FlashExplode`.
- **`PlayerHurt.Attacker` via `attacker_pawn` fallback.** `player_death` already falls back to the CS2 `attacker_pawn` handle when the attacker isn't resolvable by user ID; `player_hurt` didn't, so some hurts lost attribution for a player whose kills resolved (the "unconnected" state, #156/#172). Mirrored the fallback.
- **Deterministic collection accessors and grenade/inferno IDs.** `GrenadeProjectile`/`Inferno` `UniqueID` used `rand.Int63()` (non-reproducible per run); replaced with a monotonic counter incremented in parse order. `Participants().Playing()/Connected()/TeamMembers()/All()` were built by ranging maps (random order); now sorted by `SteamID64` then `UserID`. *Verified: identical output hash across two runs.*
- **Decode 32-bit QAngle components as noscale floats.** A QAngle field with a 32-bit component is a full-precision noscale float32, but `qangleFactory` decoded it with `readAngle(32)`, mapping raw IEEE bits into `[0,360)` and garbling the value (e.g. `m_aimPunchAngle` / `m_pAimPunchServices.m_predictableBaseAngle` reading a near-constant ~265–273°). Mirrors `floatFactory`'s noscale guard for `bitCount >= 32`. *Verified: aim-punch now sane signed degrees (±5.4° pitch); eye-angle decoding unchanged.*
- **Stop `GrenadeProjectile.Velocity()` panicking on CS2 demos.** `m_vecVelocity` is absent from the grenade projectile serializer in CS2 broadcast/GOTV, so `PropertyValueMust` panicked. Now falls back to a zero vector and documents that release kinematics come from `m_vInitialVelocity` / `m_vInitialPosition`.
- **Grenade type fallback by class + `ParserWarn` for unknown models.** `equipmentTypePerModel` is built lazily from weapon entities, so a projectile whose model hasn't been seen yet (mid-round GOTV start, post-patch variant) fell through to `EqUnknown` and printed to stderr. Now falls back to the projectile's server class and replaces the raw stderr print with a dispatched `events.ParserWarn` (`WarnTypeUnknownGrenadeModel`).

## New events / API

- **`events.InfernoFireOut`** — dispatched the first time an inferno's active fires (`m_bFireIsBurning`) transition to zero, i.e. the true flame-out (~5–7s). `InfernoExpired` fires only on entity destroy (~20s); its doc claimed the former and is corrected. *Verified: 98/98 infernos, fire-out 7.03s vs expired 20.03s median.*
- **`events.VoteCast`** — `vote_cast` fires in CS2 GOTV (not just POV, contrary to the old comment) but was unhandled. Adds `{ Player, VoteOption, Team }` with library-resolved player. *Verified: 12/12 votes resolve a player.*
- **`BombDefuseStart.Site`** — now carries the bomb site being defused, matching `BombPlanted`/`BombDefused`. *Verified: 6/6 defuse-starts carry site A/B.*
- **`PlayerJump.Position`** — populated from the player's pawn position (POV/match-server demos only, where `player_jump` fires).
- **`Player.Velocity()`** — getter wrapping `m_vecVelocity` (zero in broadcast/GOTV where it isn't networked; populated on POV/MM).
- **`Parser.FrameCount()`** — returns `-1` while the total frame count is unknown and the real count once available, so callers can branch cleanly instead of misreading `Progress()`'s `0` (the total is only backfilled from the trailing `CDemoFileInfo` at end of parse for Source 2 demos). *Verified: −1 throughout parse, real total after.*

## Documentation

- **`AmmoReserve()`** — note the era-dependent unit: the Mar-2026 CS2 reload patch changed reserve ammo to reserve magazines, so it's a magazine count on post-patch demos (AK-47 = 3, MP9 = 2), rounds before.
- **`FlashDuration`** — note it's the effective-blindness value; the engine's flash-affected window (e.g. for assist attribution) runs ~1.4× longer, with no continuous remaining-intensity netprop to expose.
- **`docs/game-events.md`** — correct CS2 broadcast-GOTV columns: `item_remove` and `player_falldamage` never fire (were ✅), `player_ping` / `player_ping_stop` do fire (were ❌); add notes that item-lifecycle events are stripped from broadcast-GOTV but recoverable via entity properties, that `bomb_beginplant`/`bomb_begindefuse` appear only in match-server SourceTV, and that the per-action event cohort isn't governed by any `tv_*` ConVar (recoverable via the entity/net-prop layer; ConVars readable via `events.ConVarsUpdated` / `GameState.ConVars()`).
- **`events.BulletDamage`** — note it and the sibling sub-tick ballistics data are recording-config gated (MM/POV/full-tick only), not on the broadcast-GOTV wire, and that advertised game-event descriptors don't imply emission.

## Notes

- Additive changes only; two behavior changes (`InfernoFireOut` events and the new `BombDefuseStart.Site`) are reflected in the regenerated `test/default.golden` — the diff contained **only** those two, no ordering/ID/flash-attribution regressions.
- The `attacker_pawn` fallback and `PlayerJump.Position` don't reproduce on complete broadcast-GOTV demos (they need POV/mid-match-start/match-server demos), so no regression test could be added from the GOTV corpus.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


markus-wa:
Rebased onto master (post-#669/#682) and pushed a follow-up commit resolving the review blockers + comments. Summary of the rebase resolutions and changes:

**Rebase resolutions**
- Master independently landed `Player.Velocity()` and `GrenadeProjectile.Velocity()` fixes — kept master's implementations, folded this branch's richer doc notes into them (the duplicated implementations dropped out as empty commits).
- The #675 `player_hurt` immediate-dispatch restructure is interleaved with the fatal-hit damage capping (both semantics preserved).
- The proto-regen commit unions with #669's usercmd protos (`CMsgServerUserCmd` incl. `DeltaData` now lives in `netmessages.pb.go`; `usercmd.proto` kept in `generate.sh`).
- `default.golden` taken from the post-#669 baseline and regenerated; **the diff to master is exactly the two behavior changes** (46× `InfernoFireOut`, 7× `BombDefuseStart.Site`) — verified line-by-line, no ordering/ID/flash-attribution regressions.

**Follow-up changes (45dbe1a)**
- **Per-parser unique ids**: `NewGrenadeProjectile`/`NewInferno`/`NewEquipment` now take an id from `parser.nextUniqueID()` instead of a package-global counter — ids stay unique within a parse and are now reproducible even under concurrent parsing (the global counter interleaved). Trade-off documented: ids from different `Parser` instances may collide, so don't key cross-demo data by them. Also removes the order-dependency when several demos parse in one process.
- `bombBeginDefuse` no longer dereferences `Bomb()` unguarded when deriving the site (mid-round parse start edge).
- `processInfernoFireOut` reads only the `m_bFireIsBurning` flags instead of rebuilding the full fire list per inferno per frame (with a note that it's intentionally not gated by `DisableMimicSource1Events`).
- Grenade-model warnings kept as a switch with `//nolint:exhaustive` (staticcheck QF1003 dislikes the if/else form, so the two linters cancel out — the switch + nolint matches the pattern used elsewhere since the linter upgrade).
- Dropped stale `//nolint:gosec` directives (G115 is excluded globally now) and fixed the new `wsl_v5`/`funlen`/`gocognit` findings.

Build, vet, full `./pkg/...` + `./examples/...` suites, race on the demo test — all green locally.

markus-wa:
### Note: flashbang attribution grace-window

The refactored flashbang pairing (entity-keyed detonation marking + per-frame victim buffer) removes the one-frame grace window the old FIFO code had: previously a pawn whose `m_flFlashDuration` update arrived the frame *after* detonation was still attributed to the (still-head) flashbang and dispatched; now victims are only attributed on the exact detonation frame, so a late-arriving update is silently dropped (`detonated` is empty on the next frame).

Verified same-frame arrival on the 13 broadcast-GOTV demos, so this is a non-issue there — but worth a spot-check on a POV / match-server (SourceTV) demo, where game-event/prop-update ordering can differ, in case it drops `PlayerFlashed` events. A one-line comment noting the same-frame assumption would help future readers.

markus-wa:
### Note: exported constructor signatures are a breaking change

`NewEquipment(wep, id)`, `NewGrenadeProjectile(id)` and `NewInferno(..., id)` are exported functions in the `common` package and their signatures changed in the final commit. They're documented `Intended for internal use only`, but any external caller that uses them directly will fail to compile.

CI's `test-csda` job passes, so the known downstream consumer is unaffected — but per semver this is a breaking API change for the v5 line. Probably warrants a major version bump (v6) before merging, or at minimum a prominent entry in the release notes.

Issue #340, opened 2022-04-06
Get player's button presses infomation


hx-w:
**Higher goal**
I want to get player's button presses infomation in a tick frame, so I can know whether player is pressing `W A S D` or not (this is not meaning player is really moving front, left, back or right).

In sourcemod, player's button info will be encoded with [different button presses](https://github.com/alliedmodders/sourcemod/blob/01203a5a449b5e8a26e6f3fa6b91a3a8b9a68c07/plugins/include/entity_prop_stocks.inc#L99-L125) and I can get player's button by using function [`int GetClientButtons(int client)`](https://github.com/alliedmodders/sourcemod/blob/01203a5a449b5e8a26e6f3fa6b91a3a8b9a68c07/plugins/include/entity_prop_stocks.inc#L574-L594).

I noticed that there is no such method to get player's button presses in [Player](https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v2@v2.13.1/pkg/demoinfocs/common#Player), and I tried with using     `player.Entity.PropertyValue("m_nButtons")` and got false, since there is no `"m_nButtons"` property in `player.Entity.ServerClass()`. So I wonder if there could be a method to implement this feature.

**Describe the solution you'd like**

It will be very nice if there is a `Buttons()` or `ButtonsPressing()` method which return a `int32` value that can be decoded in a specific way in [Player](https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v2@v2.13.1/pkg/demoinfocs/common#Player). **Or** just tell me how to get `"m_nButtons"` or some properties relevant to button pressing else. 

**Describe alternatives you've considered**

There is a callback function [OnPlayerRunCmd](https://sm.alliedmods.net/new-api/sdktools_hooks/OnPlayerRunCmd) in sourcemod, and parameter `int& buttons` is what I metioned above.

If it is difficult to get button presses, I wish I can get player's **desired velocity** (`float vel[3]`) instead (not [actual velocity](https://pkg.go.dev/github.com/markus-wa/demoinfocs-golang/v2@v2.5.0/pkg/demoinfocs/common#Player.Velocity)).

> Player's desired velocity is presented as {450, 0, 0} when player pressing `W` even if he is not actually moving front.
> And {-450, 0, 0} when pressing `S`, {0, -450, 0} when pressing `A`, {0, 450, 0} when pressing `D` for example.
[...]

PR #669, opened 2026-08-19
fix: fully decode delta-encoded CS2 user commands

WangChuDi:
## Description

PR #661 restores button extraction for CS2 demos after the 2026-07-09 update, but its decoder keeps only one mutable snapshot per player slot and skips the indexed repeated fields used by Valve's delta encoder. This PR keeps that button behavior and completes the game-compatible reconstruction path.

This branch includes the unmerged #661 commit as a prerequisite, then adds the decoder changes described below.

## Changes

- Replace the single per-slot snapshot with the 150-slot command-number ring used by client.dll, including exact command-number validation and baseline-mismatch rejection.
- Decode wire type 7 as a protobuf-default reset marker for scalar, nested, and repeated fields.
- Recursively merge nested messages instead of treating their payload as an opaque replacement.
- Decode Valve's indexed repeated operation stream for subtick_moves and input_history, including index-zero patches, wire-2 element updates, wire-7 list-length changes, and malformed-operation rejection.
- Ignore usercmds embedded in DEM_FullPacket checkpoints because they are seek snapshots rather than transports delivered to the command handler.
- Expose every successfully reconstructed command through events.UserCmd, carrying the full generated CSGOUserCmdPB plus player slot, command number, server tick, and client tick metadata.
- Add UserCmdParsingFull (the default), UserCmdParsingButtonsOnly, and UserCmdParsingDisabled. ButtonsOnly follows the PR #661 low-overhead buttonstate1 path while retaining the 150-slot command-number validation; Disabled skips user-command reconstruction and leaves legacy property fallback available.
- Preserve PlayerButtonsStateUpdate and the legacy button-property fallback for demos that do not contain CSVCMsg_UserCommands.

## Public API

Consumers can register an events.UserCmd handler and read all reconstructed fields from event.Command, including base.buttons_pb.buttonstate1/2/3, mouse deltas, weapon selection, view angles, movement, subtick moves, input history, prediction flags, and execution notes. ParserConfig.UserCmdParsing defaults to UserCmdParsingFull; applications that only need PlayerButtonsStateUpdate can select UserCmdParsingButtonsOnly, and applications that do not need user-command input can select UserCmdParsingDisabled.

### Using UserCmdParsingButtonsOnly

ButtonsOnly is an explicit opt-in. It follows the `base.buttons_pb.buttonstate1` path without `proto.Unmarshal`, full protobuf snapshot allocation, or repeated-field reconstruction. It preserves the 150-slot command-number baseline checks and emits `events.PlayerButtonsStateUpdate`; it intentionally does not emit `events.UserCmd`, because the other command fields are not reconstructed.

```go
config := demoinfocs.DefaultParserConfig
config.UserCmdParsing = demoinfocs.UserCmdParsingButtonsOnly

err := demoinfocs.ParseFileWithConfig("match.dem", config, func(p demoinfocs.Parser) error {
	 p.RegisterEventHandler(func(e events.PlayerButtonsStateUpdate) {
		 fmt.Println(e.Player, e.ButtonsState)
	 })
	 return nil
})
```

Use `UserCmdParsingFull` (the default) and register `events.UserCmd` when mouse, weapon, view, movement, subtick, input-history, or other full command fields are required. Use `UserCmdParsingDisabled` when neither usercmd events nor button-state reconstruction is needed; the outer `CSVCMsg_UserCommands` message still has to be read by the generic message parser, but no usercmd payload reconstruction is performed.

### Fields affected by delta_data

`delta_data` is a delta representation of the existing `CSGOUserCmdPB` command, not a buttons-only payload. In full mode, the decoder can update every field in these message groups:

- `CSGOUserCmdPB`: `base`, repeated `input_history`, `attack1_start_history_index`, `attack2_start_history_index`, `left_hand_desired`, `is_predicting_body_shot_fx`, `is_predicting_head_shot_fx`, and `is_predicting_kill_ragdolls`.
- `CBaseUserCmdPB`: `legacy_command_number`, `client_tick`, `prediction_offset_ticks_x256`, `buttons_pb.buttonstate1/2/3`, `viewangles`, `forwardmove`, `leftmove`, `upmove`, `impulse`, `weaponselect`, `random_seed`, `mousedx`, `mousedy`, `pawn_entity_handle`, `move_crc`, `consumed_server_angle_changes`, `cmd_flags`, and `execution_notes.ignored_reason`.
- `CSubtickMoveStep` repeated entries: `button`, `pressed`, `when`, `analog_forward_delta`, `analog_left_delta`, `pitch_delta`, and `yaw_delta`.
- `CSGOInputHistoryEntryPB` repeated entries: view angles, render/player tick and fraction fields, client/server/player interpolation data, frame number, target entity index, shoot position, target head/absolute position checks, and target absolute angle checks.
- `CMsgServerUserCmd` transport metadata: player slot, command number, server tick executed, client tick, full `data`, and delta-encoded `delta_data`.

ButtonsOnly intentionally reads only `base.buttons_pb.buttonstate1`; all other fields above are skipped for performance. Full mode retains their inherited baseline values, applies wire-type-7 resets, merges nested messages, and applies indexed repeated deltas.

### Newly added library fields and API

The game schema already defines the command fields above. This PR adds or exposes the following library-side items:

- Generated `CMsgServerUserCmd.DeltaData` support for field 6, while retaining the legacy `Data` path.
- Public `events.UserCmd` with `Player`, `PlayerSlot`, `CommandNumber`, `ServerTickExecuted`, `ClientTick`, and the complete reconstructed `Command`.
- `ParserConfig.UserCmdParsing` with `UserCmdParsingFull`, `UserCmdParsingButtonsOnly`, and `UserCmdParsingDisabled`.
- `WarnTypeUserCommandBaselineMissing` and `WarnTypeUserCommandBaselineMismatch` for observable baseline failures.

## Validation

The real test demo and all performance measurements use the exact sample attached in [demoparser issue #340 comment 4995197071](https://github.com/LaihoE/demoparser/issues/340#issuecomment-4995197071): the [Google Drive demo](https://drive.google.com/file/d/1TQfpcwqvi52BAn-wOVxhykJdQMxhcVPK/view?usp=sharing), downloaded locally as issue340_pr343_sample.dem.

### Game DLL oracle comparison

The game-side reference is the public [CS2 usercmd DLL oracle](https://github.com/WangChuDi/cs2-usercmd-dll-oracle). It loads an instrumented probe into an offline `-insecure` CS2 process through HLAE/AfxHookSource2. IDA was used to identify the usercmd handler, the pre-decode transport point, and the post-decode ring/cache update point. The probe records transport metadata before the handler, captures the command after client.dll has reconstructed and stored it, then serializes the game's resulting CSGOUserCmdPB. It does not replace the game's decoder and is not a runtime dependency of this parser.

Each oracle record is classified as Decoded or Rejected. Decoded records contain the complete post-decode payload. Rejected records are transports for which client.dll did not reach the post-decode point, so there is no complete payload to export; the parser does not invent one or emit an events.UserCmd event for it.

| Item | Game DLL oracle | Current demoinfocs full mode | Difference |
| --- | ---: | ---: | ---: |
| Transport records | 1,323,304 | 1,323,304 aligned* | 0 |
| Full `data` records | 2,221 | 2,221 aligned* | 0 |
| `delta_data` records | 1,321,083 | 1,321,083 aligned* | 0 |
| Successfully reconstructed payloads | 1,302,947 | 1,302,947 `events.UserCmd` | 0 |
| Rejected records | 20,357 | 20,357 aligned; no `events.UserCmd` | 0 |
| Oracle write failures | 0 | N/A | N/A |
| Player slots | 1-10 | 1-10 | 0 |
| Records with non-zero buttons | 841,739 | 841,739 | 0 |
| Records with mouse movement | 313,180 | 313,180 | 0 |
| Weapon-selection records | 13,304 | 13,304 | 0 |
| Records containing subtick moves | 59,713 | 59,713 | 0 |
| Maximum subtick moves in one command | 7 | 7 | 0 |
| Maximum input-history entries | 5 | 5 | 0 |
| `cannot_move` | 261,665 | 261,665 | 0 |
| `dead` | Not present in the DLL payload | Not a `CSGOUserCmdPB` field | N/A |

The `*` transport rows are alignment-scope results, not public event counts: `events.UserCmd` emits successfully reconstructed payloads and rejected transports are intentionally not emitted. A direct full-mode run produced the event-level values in the table. The streaming differential harness matched all 1,302,947 decoded payloads and 20,357 rejected outcomes, with zero protobuf field differences across scalar fields, nested messages, repeated input-history entries, repeated subtick entries, command metadata, and ring/cache command numbers. The production parser does not load client.dll, require HLAE, or depend on the oracle repository.

### PlayerButtons-only public event comparison

This section compares only the PlayerButtons output: the public `PlayerButtonsStateUpdate` event sequence containing `(Player.UserID, ButtonsState)`. It does not compare the complete `events.UserCmd` payload or mouse movement, weapon selection, view angles, movement, subtick moves, input history, or other usercmd fields. The raw usercmd replay comparison and this button-only public-event comparison measure different layers. The raw replay compares reconstructed button payloads for each transport; this test executes each parser and compares the actual `PlayerButtonsStateUpdate` sequence. The Full parser is used only as the button-state reference; the current ring before validation, the current validation-only implementation, and `599c25e` are run in their button-only paths. Because `PlayerButtonsStateUpdate` does not expose a command number, events are aligned per player slot by `(Player.UserID, ButtonsState)` while preserving order, with `MsgQueueBufferSize=0`.

| Player slot | Full parser PlayerButtonsState events | Current ring (pre-validation) ButtonsOnly extra events | Current ButtonsOnly (validation-only) extra events | PR #661 commit 599c25e ButtonsOnly extra events |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 6,115 | 0 | 0 | 27 |
| 2 | 7,345 | 0 | 0 | 31 |
| 3 | 5,858 | 2 | 0 | 198 |
| 4 | 6,643 | 0 | 0 | 159 |
| 5 | 7,007 | 0 | 0 | 151 |
| 6 | 7,691 | 0 | 0 | 35 |
| 7 | 4,531 | 0 | 0 | 166 |
| 8 | 7,444 | 0 | 0 | 133 |
| 9 | 7,855 | 1 | 0 | 227 |
| 10 | 6,976 | 2 | 0 | 165 |
| **Total** | **67,465** | **5** | **0** | **1,292** |

All 67,465 full-parser `PlayerButtonsStateUpdate` events were found in all three button-only candidate sequences with the same player slot and button value. Neither the pre-validation ring nor the validation-only sequence had a changed value for a matched event. The pre-validation current ring inserted five additional public updates: button states `8` and `0` for slot 3, `514` for slot 9, and `10` and `1026` for slot 10. The validation-only sequence inserted no additional public updates. The 599c25e implementation inserted 1,292 additional updates across all slots. Relative to the full decoder, the pre-validation ring and 599c25e updates are false positives with count deviations of 0.007411% and 1.915067%, while validation-only has a 0% deviation.

This result does not claim that the two implementations are equivalent. The earlier raw replay row reused the current button scanner while simulating only 599c25e's single-value baseline policy; it did not execute the old parser's complete event-dispatch path. The event-level comparison above runs the actual 599c25e parser and demonstrates that its public output differs materially from the full decoder. This is a parser-to-parser public API comparison; the DLL oracle exports reconstructed usercmd payloads, not demoinfocs event objects.

Focused verification passed with go test ./pkg/demoinfocs -run TestApplyUserCmdDelta -count=1, go test ./pkg/demoinfocs -run TestUserCmdRing -count=1, go test ./pkg/demoinfocs -run TestServerUserCmdDeltaDataRoundTrip -count=1, the button-only decoder tests, and the affected events/fake/msg/sendtables packages.

The repository-wide go test ./... command could not complete in this checkout because the standard example fixture test/cs-demos/s2/s2.dem is absent and the environment cannot write the global Go module-cache lock files; these are checkout/environment issues, not test failures in the changed packages.

## Performance

The benchmark compares fork origin/master at ed22bca with this PR at b0612bc using the exact issue #340 demo above, Windows amd64, Go go1.25.6, AMD Ryzen 5 5600, GOMAXPROCS=1, and five repetitions. The demo bytes are loaded before the timer; each iteration measures NewParser, ParseToEnd, and Close. The table reports the median of the five runs.

| Mode | Parse time | Allocated memory | Allocations |
| --- | ---: | ---: | ---: |
| origin/master | 6,293,693,700 ns/op | 1,925,297,288 B/op | 25,374,790 allocs/op |
| This PR: UserCmdParsingButtonsOnly | 7,090,641,500 ns/op | 1,928,159,136 B/op | 25,500,993 allocs/op |
| This PR: UserCmdParsingFull | 19,289,515,600 ns/op | 6,324,277,048 B/op | 183,497,175 allocs/op |
| This PR: UserCmdParsingButtonsOnly (validation-only) | 7,541,456,500 ns/op | 1,928,225,120 B/op | 25,501,113 allocs/op |

The existing UserCmdParsingButtonsOnly row is intentionally retained as the non-validation baseline. The additional validation-only row is a fresh five-run median using the current validation-only implementation, the same issue #340 demo, GOMAXPROCS=1, and NewParser, ParseToEnd, and Close per iteration. It validates all delta fields while retaining only buttons and repeated-list lengths. Compared with this fresh Full run, validation-only ButtonsOnly reduces parse time by 56.64%, allocated memory by 69.51%, and allocations by 86.10%.

Relative to origin/master, ButtonsOnly adds +12.66% parse time, +0.15% allocated memory, and +0.50% allocations. Relative to the full decoder, ButtonsOnly reduces parse time by 63.24%, allocated memory by 69.51%, and allocations by 86.10%. Full decoding adds +206.49% parse time, +228.48% allocated memory, and +623.15% allocations because it creates complete protobuf snapshots for buttons, mouse, weapon selection, nested fields, subtick moves, and input history. The default remains full parsing as requested in PR #661; the benchmark makes the opt-in trade-off explicit.


markus-wa:
Maintainer-reviewed and pushed two commits to this branch:

**1. Fix: resolve the player via controller entity, not userID** (4db112c)

The handlers looked the player up with `playersByUserID[int(slot)]`, but `playersByUserID` is keyed by `UserID` (from the player-info string table) while `CMsgServerUserCmd.player_slot` is a player slot (controller-entity space). Those namespaces only coincide by accident; on demos with reconnects or different join orders the button events misattributed to the wrong player or were silently dropped.

Now resolved via `playersByEntityID[int(slot)+1]` in both `handleUserCommands` and `handleUserCommandButtons` — slot N maps to controller entity N+1, the same convention `getOrCreatePlayerFromControllerEntity` already relies on (`rawPlayers[controllerEntityID-1]`).

Since the existing validation compared parser-vs-parser (same lookup on both sides) and the DLL oracle was transport/slot-level, this wasn't caught — worth re-running the oracle differential on a demo with reconnects to confirm. I understand you may be able to verify this empirically.

**2. Default `UserCmdParsingButtonsOnly`, not Full** (b509597)

Full decoding is expensive (+206% parse time / +228% mem / +623% allocs on the reference demo) and the reconstructed `events.UserCmd` payloads are a new API, while ButtonsOnly restores exactly the pre-update public behavior (`PlayerButtonsStateUpdate`, 0% event deviation) at a fraction of the cost. Full stays available via `ParserConfig.UserCmdParsing = UserCmdParsingFull`.

To keep `ParserConfig{}` consistent with `DefaultParserConfig`, the enum was reordered so `UserCmdParsingButtonsOnly` is the zero value.

Full `./pkg/...` suite passes unchanged (incl. the golden demo test), race-clean on the usercmd tests.

PR #693, opened 2026-09-02
fix: keep reconnect name changes & dedupe re-used bot slots

### LINKED ISSUE #692: v6.0.0-alpha.0: name change reverted after reconnect; Playing() returns the same *Player twice after a bot name is reused
Two player-identity regressions in `v6.0.0-alpha.0` relative to `v5.2.0`, found while upgrading our production CS2 parser to v6. They surfaced on the same demo and may share a root area, so they are reported together — happy to split into two issues if you prefer.

**Environment:** `github.com/markus-wa/demoinfocs-golang/v6 v6.0.0-alpha.0`, go1.27.0, windows/amd64.

**Demo (shows both):** a 197 MB CS2 GOTV demo from a real match, shared for this report: https://drive.google.com/file/d/1M1J1w61yTCGPnRuSNSVTGm7DP2HtSIUC/view?usp=sharing
The file is **gzip-compressed** (`1f 8b` magic) — gunzip first, or use the repro programs below, which auto-detect it.

Both behaviours were also swept across our 52-demo production corpus; the numbers below say how common each is. `v5.2.0` shows neither on any of the 52.

---

## 1. A reconnecting player's name change is reverted a few ticks later

Player 4 ("ZoRoBeast") disconnects at tick 17956 and reconnects at tick 19183 under a new name. The server's own `player_connect` game event carries the new name. v6 dispatches `PlayerNameChange` to the new name and then, four ticks later, dispatches a change **back** to the old name, which then sticks — `Player.Name` reports the pre-reconnect name for the rest of the match. The same demo has a second instance (player 9 at ticks 11370/11375). Repro output on the demo above:

```
tick  11370  PlayerNameChange  uid=9 "iBaaaaaaad" -> "BuddhaMonsteRRR"
tick  11370  player_connect    userid=9 name="BuddhaMonsteRRR"
tick  11375  PlayerNameChange  uid=9 "BuddhaMonsteRRR" -> "iBaaaaaaad"
tick  19183  PlayerNameChange  uid=4 "ZoRoBeast" -> "- [Z]oRo"
tick  19183  player_connect    userid=4 name="- [Z]oRo"
tick  19187  PlayerNameChange  uid=4 "- [Z]oRo" -> "ZoRoBeast"
```

`v5.2.0` on the same demo dispatches the uid=4 change once and keeps the new name (it never dispatches anything for uid=9). Corpus: **15 of 52 demos show this, 23 reverts in total, every one 2–21 ticks after a change that coincides with a disconnect→reconnect under a new name.**

One observation that may help locate it: in both v5 and v6, the controller entity's `m_iszPlayerName` stays on the *old* name for the whole match, so the late revert presumably comes from a string-table entry being re-applied rather than from the entity.

<details>
<summary>Repro program (imports only demoinfocs)</summary>

```go
// go run . <demo.dem>   (gzip or plain)
package main

import (
	"compress/gzip"
	"fmt"
	"io"
	"os"

	dem "github.com/markus-wa/demoinfocs-golang/v6/pkg/demoinfocs"
	"github.com/markus-wa/demoinfocs-golang/v6/pkg/demoinfocs/events"
)

func main() {
	f, err := os.Open(os.Args[1])
	if err != nil {
		panic(err)
	}
	defer f.Close()

	var r io.Reader = f
	if zr, err := gzip.NewReader(f); err == nil {
		r = zr
	} else if _, err := f.Seek(0, io.SeekStart); err != nil {
		panic(err)
	}

	p := dem.NewParser(r)
	defer p.Close()

	p.RegisterEventHandler(func(e events.GenericGameEvent) {
		if e.Name == "player_connect" {
			fmt.Printf("tick %6d  player_connect    userid=%d name=%q\n",
				p.GameState().IngameTick(), e.Data["userid"].GetValShort(), e.Data["name"].GetValString())
		}
	})

	p.RegisterEventHandler(func(e events.PlayerNameChange) {
		fmt.Printf("tick %6d  PlayerNameChange  uid=%d %q -> %q\n",
			p.GameState().IngameTick(), e.Player.UserID, e.OldName, e.NewName)
	})

	if err := p.ParseToEnd(); err != nil {
		panic(err)
	}
}
```

</details>

---

## 2. `Participants().Playing()` returns the same `*Player` twice after a bot slot's name is reused

In this demo bot "Ava" is user id 11 (leaves at tick 19340) and, later, a bot also named "Ava" is user id 12 (ticks 32350–34844). In roughly **4 parses out of 10**, `Playing()` returns the **same `*Player` pointer twice** for the second bot's entire lifetime (2,494 ticks) — one object registered under both user-id keys. Which id the object carries varies from parse to parse:

```
tick 32350: uid 12 twice — "Ava" ent=13 and "Ava" ent=13, same object: true
tick 32350: uid 11 twice — "Ava" ent=13 and "Ava" ent=13, same object: true
0 tick(s) with a duplicate user id            <- the other parses
```

`v5.2.0` printed `0 tick(s)` in 21 of 21 parses of this demo. The run-to-run variance smells like map iteration order; a hypothesis from reading `getOrCreatePlayerFromControllerEntity`: bots are matched to a `rawPlayers` entry **by name** while ranging over the map, this demo has two entries named "Ava", and when iteration lands on the old bot's entry its existing `Player` is reused for the new controller and registered under a second user-id key without the first being dropped — so `playersByUserID` holds one object under two keys and every accessor built from it yields it twice.

For consumers this means duplicate per-tick rows for that user id on some parses and not others — and when the object keeps the *old* id, the second bot's data is silently attributed to the first bot.

<details>
<summary>Repro program (imports only demoinfocs; run it a few times)</summary>

```go
// go run . <demo.dem>   (gzip or plain; nondeterministic — run a few times)
package main

import (
	"compress/gzip"
	"fmt"
	"io"
	"os"

	dem "github.com/markus-wa/demoinfocs-golang/v6/pkg/demoinfocs"
	"github.com/markus-wa/demoinfocs-golang/v6/pkg/demoinfocs/common"
	"github.com/markus-wa/demoinfocs-golang/v6/pkg/demoinfocs/events"
)

func main() {
	f, err := os.Open(os.Args[1])
	if err != nil {
		panic(err)
	}
	defer f.Close()

	var r io.Reader = f
	if zr, err := gzip.NewReader(f); err == nil {
		r = zr
	} else if _, err := f.Seek(0, io.SeekStart); err != nil {
		panic(err)
	}

	p := dem.NewParser(r)
	defer p.Close()

	seen := map[int]*common.Player{}
	last, dupTicks := -1, 0

	p.RegisterEventHandler(func(events.FrameDone) {
		t := p.GameState().IngameTick()
		if t == last {
			return
		}

		last = t
		clear(seen)

		for _, pl := range p.GameState().Participants().Playing() {
			other, dup := seen[pl.UserID]
			seen[pl.UserID] = pl

			if !dup {
				continue
			}

			if dupTicks == 0 {
				fmt.Printf("tick %d: uid %d twice — %q ent=%d and %q ent=%d, same object: %v\n",
					t, pl.UserID, other.Name, other.EntityID, pl.Name, pl.EntityID, other == pl)
			}

			dupTicks++
		}
	})

	if err := p.ParseToEnd(); err != nil {
		panic(err)
	}

	fmt.Printf("%d tick(s) with a duplicate user id\n", dupTicks)
}
```

</details>

---

To run either program:

```
mkdir repro && cd repro && go mod init repro
go get github.com/markus-wa/demoinfocs-golang/v6@v6.0.0-alpha.0
# save a program above as main.go
go run . path/to/demo.dem
```

Very happy to test a branch or candidate fix against our 52-demo production corpus — and thanks for #661/#669; restoring the button state from user commands is the reason we are moving to v6.

markus-wa:
Closes #692

Two player-identity regressions on CS2 GOTV demos, both rooted in **stale userinfo string-table state for player slots that were re-used after a disconnect/reconnect under a new name**.

### 1. Name change reverted a few ticks after reconnect
The GOTV proxy keeps the *pre-reconnect* userinfo for the re-used slot and re-sends it forever — verified as `CSVCMsg_UpdateStringTable` single-entry updates arriving +4 ticks after the rename and then periodically every 3840/7680 ticks (`CDemoStringTables` state dumps), plus one `svc_CreateStringTable` full re-send. `parseUserInfo` re-applied the stale name on every one of them.

v5.2.0 was immune because `parseUserInfo` had a `if _, exists := p.rawPlayers[playerIndex]; exists { return }` guard (from f111430, 2023). It was removed in 44984e7/22c069f ("broadcasts: fix player user-id mappings when starting parsing mid-match", May 2026, first shipped in v6.0.0-alpha.0) — that fix is preserved here because the signon table (`CSVCMsg_CreateStringTable`, 13 tables at tick 0) is the *first* apply per slot, which the guard always passes.

### 2. `Playing()` returning the same `*Player` twice after a bot name is re-used
`getOrCreatePlayerFromControllerEntity` matched bot controllers to `rawPlayers` **by name while ranging over the map** (also 44984e7). With two entries named "Ava" (stale + current), map iteration order randomly picked the stale entry, and `getOrCreatePlayer` re-stamped `player.UserID` from it and registered the object under a second user-ID key — one object under two keys, nondeterministically (~4/10 parses, alternating which ID the object carried).

## Fixes

1. **`stringtables.go`** — restore the existence-guard, scoped to entries *re-applying* info for a slot we already have raw info for. The signon table and disconnects clearing slots keep reconnect/rename applies working; broadcast mid-match mapping (the reason the guard was removed) keeps working because full-table first applies always pass.
2. **`game_events.go`** — `player_connect` (which carries the authoritative name for the new session) now syncs onto the player object, dispatches `PlayerNameChange`, and records the session info under the slot key so the guard suppresses the proxy's stale state for that slot. This also fixes the same-tick reconnect race (uid 9: v5 showed the old name all match).
3. **`datatables.go`** — bot controller → raw info matching is slot-first (slot N = entity N+1, the convention already relied on by `s2_commands.go` and the old `rawPlayers[controllerEntityID-1]` fallback) with the controller's props cross-checked; the name/XUID scan remains as a fallback for broadcast drift but iterates slots in ascending order so it's deterministic.
4. **`datatables.go`** — the `m_iConnected` disconnection handler deleted *every* `rawPlayers` entry matching the disconnecting player's XUID, including a reconnecting player's fresh same-frame entry, re-arming the guard mid-frame. It now only drops the disconnected session's entry (user-ID encoded without the slot byte — the legacy-userid encoding from 919a289).

## Verification

Reported demo (`~/Downloads/broken_1.dem`, 10+ parses, all deterministic):

| | before (v6.0.0-alpha.0) | after |
|---|---|---|
| uid 4 rename at 19183 | applied, reverted at 19187 & periodically (19201, 23041, 30721, 34561, 38401, …) | applied once, **stuck** |
| uid 9 rename at 11370 | applied, reverted at 11375 (v5 showed the old name all match) | applied at 11370, **stuck** |
| duplicate `Playing()` user-ID | ~4/10 parses, 2,494 ticks each | **0 in 10/10 parses** |

Regression tests (`issue692_test.go`) read `broken_1.dem` from the `cs-demos` submodule and skip when it isn't initialized. The demo was added to the submodule as 90MiB gzip parts (`cat s2/broken_1.dem.gz.part.aa s2/broken_1.dem.gz.part.ab s2/broken_1.dem.gz.part.ac > s2/broken_1.dem.gz && gunzip s2/broken_1.dem.gz`) since GitLab rejects >100MiB blobs.

- `go test ./... -short` ✅
- `go test ./pkg/demoinfocs/ -run Test692 -count=1` (with submodule) ✅ 5× runs
- `gofmt -l pkg/` ✅, `go vet ./...` ✅

Happy to test against your 52-demo production corpus feedback — the guard's per-session semantics are exactly v5's, which showed neither behaviour on any of the 52.


markus-wa:
**CI failure explained & fixed** (c1f91b6): the golden-file mismatch in `TestDemoInfoCs` is expected fallout from restoring the existence-guard.

The golden was regenerated on Aug 31 — *after* the guard removal in 44984e7 — so it contains **77 events** from periodic GOTV userinfo dumps re-applying **identical** info to existing slots (`PlayerInfo` + `StringTablePlayerUpdateApplied` with unchanged names; zero `PlayerNameChange`, no connect/disconnect in between, verified by diffing the decoded golden against the actual output). Restoring v5's guard suppresses exactly those idempotent re-applies.

Verified before regenerating: the golden↔actual diff is **purely those 77 deletions with zero additions** — no behavioural event was dropped. Regenerated the golden with `-update`, and the full suite (incl. `Test692*` against the submodule demo) is green locally.

PR: 5/10 parses previously showed the duplicate-`*Player` bug; the golden's uid 9 reconnect apply (`Index:9` re-apply + `Index:65289`) in `s2.dem` is also now handled by the connect-event path, matching the #692 demo behaviour.

